PR #187, opened 2016-07-27
Improve user synchronization with Memberful

mrhead:
Previously we were updating only user's first name, last name and email address. But we did not touch nickname and display name. Some customers actually use it, so it is better when we upgrade it every time we sync user with Memberful.


mrhead:
We decided not to do this because it would affect all users and this is what something what users can handle by themselves.

PR #200, opened 2017-05-10
CSS class for the links separator.

dstrojny:
@mrhead Please check and confirm this is done properly. Let's **not** release it immediately. We can include it with our next plugin update.

### DIFF
--- a/wordpress/wp-content/plugins/memberful-wp/src/widgets.php
+++ b/wordpress/wp-content/plugins/memberful-wp/src/widgets.php
@@ -124,7 +124,7 @@ function memberful_wp_format_widget_links( $links ) {
     $links[$key] = '<a href="'.$link['href'].'" class="'.$link['class'].'">'.$link['text'].'</a>';
   }
 
-  return implode( ' | ', $links );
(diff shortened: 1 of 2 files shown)


wordpress/wp-content/plugins/memberful-wp/src/widgets.php
```diff
@@ -124,7 +124,7 @@ function memberful_wp_format_widget_links( $links ) {
     $links[$key] = '<a href="'.$link['href'].'" class="'.$link['class'].'">'.$link['text'].'</a>';
   }
 
-  return implode( ' | ', $links );
+  return implode( ' <span class="memberful-links-separator">|</span> ', $links );
```
mrhead: This class isn't defined anywhere in the plugin? Is this intentional and do we expect our users to define it if they want to change things?

dstrojny: Yes, we had a customer who wanted to hide the separator with CSS because they were styling the widget themselves. This class makes that possible.

PR #201, opened 2017-05-19
Add support for past due subscriptions

mrhead:
We have implemented multi-day renewals recently. Basically it means that
a subscription can be active up to two days after the expiration date.
Therefore it is not enough to check the `expires_at` attribute to find
out if subscription is active or not. The best way how to check
subscription status is to trust Memberful and check only the `active`
attribute.

Unfortunately we can't rely only on the `active` attribute because we
weren't storing it before (and our API wasn't providing it). Therefore,
first we will check if the `active` attribute is set. If yes, then we
will use it. If not, then we fallback back to the `expires_at` attribute.


mrhead:
Closing this in favor of #202.

PR #208, opened 2017-08-11
Cleanup memberful_wp_cron_sync_users

eric:
It appears that $mapper is no longer used in memberful_wp_cron_sync_users().


mrhead:
You are right, thanks!

PR #214, opened 2017-11-14
Better existing WP user account sync flow

mrhead:


wordpress/wp-content/plugins/memberful-wp/src/authenticator.php
```diff
@@ -307,7 +307,8 @@ function memberful_wp_add_nonce_check_to_login_form() {
 
 function memberful_wp_display_check_account_message() {
   if ( isset($_GET['memberful_account_check']) ) {
-    return '<p>'.__( 'We found an existing WordPress user account with your email address. Please sign in so we can sync the accounts.' ).'</p>';
+    $message = apply_filters( 'memberful_account_check_message', __( 'We found an existing account with the same email address. To confirm you own the account, please sign in with your pre-existing password.' ) );
```
mrhead: @dstrojny The new filter name is `memberful_account_check_message`. Please let me know if it is ok or if you want to change it.

dstrojny: @mrhead That's good!

PR #217, opened 2017-12-04
Modify overlay script

mrhead:


wordpress/wp-content/plugins/memberful-wp/views/embed.js.php
```diff
@@ -17,11 +17,7 @@
 
     setup = function() { window.MemberfulEmbedded.setup(); }
 
-    if(s.addEventListener) {
-      s.addEventListener("load", setup, false);
-    } else {
-      s.attachEvent("onload", setup);
-    }
+    s.addEventListener("load", setup, false);
```
mrhead: This change is ok and we already use the same embed code in the admin dashboard. However, other changes in this PR should be discarded before we merge this.

mrhead:
@lmansur @northerner This is an old pull request, but I think that it is still adequate and good to go. Could you please review it?

PR #220, opened 2018-02-04
Basic WooCommerce SUpport

LiamBailey:
Hides add to cart button from archive and single pages and shows marketing content in its place. Blocks adding to cart anyway.


mrhead:
Thank you!

PR #222, opened 2018-02-09
Fix products getting swept up in global the_content protection

LiamBailey:
Pulled in latest changes from main fork and then pushed this fix on top

### DIFF
--- a/wordpress/wp-content/plugins/memberful-wp/src/contrib/woocommerce.php
+++ b/wordpress/wp-content/plugins/memberful-wp/src/contrib/woocommerce.php
@@ -38,7 +38,7 @@ class Memberful_Wp_Integration_WooCommerce {
    * Archive (loop) pages
    */
   function is_purchasable( $purchasable, $product ) {
-    return !current_user_can('administrator') && !memberful_can_user_access_post( get_current_user_id(), $product->get_id() ) ? false : $purchasable;
+    return !memberful_is_admin( wp_get_current_user() ) && !memberful_can_user_access_post( get_current_user_id(), $product->get_id() ) ? false : $purchasable;
   }
 
   /**
@@ -47,7 +47,7 @@ class Memberful_Wp_Integration_WooCommerce {
   function hide_add_to_cart_button() {
     global $post;
 
-    if (!current_user_can('administrator') && !memberful_can_user_access_post( get_current_user_id(), $post->ID ) ) {
+    if ( !memberful_is_admin( wp_get_current_user() ) && !memberful_can_user_access_post( get_current_user_id(), $post->ID ) ) {
       remove_action( 'woocommerce_single_product_summary', 'woocommerce_template_single_add_to_cart', 30 );
       echo $this->memberful_wp_protect_woo_content( $post->ID );
     }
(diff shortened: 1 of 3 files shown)


wordpress/wp-content/plugins/memberful-wp/src/contrib/woocommerce.php
```diff
@@ -18,15 +18,27 @@ public static function instance() {
...
   }
 
   /**
    * Makes it so that the add to cart button on archive page will show Read More and link to product on
    * Archive (loop) pages
    */
-  function not_purchasable( $purchasable, $product ) {
-    return !memberful_can_user_access_post( get_current_user_id(), $product->get_id() ) ? false : $purchasable;
+  function is_purchasable( $purchasable, $product ) {
+    return !current_user_can('administrator') && !memberful_can_user_access_post( get_current_user_id(), $product->get_id() ) ? false : $purchasable;
```
mrhead: `current_user_can("administrator")` doesn't work well with multisite mode according to [this stack overflow question](https://wordpress.stackexchange.com/questions/199694/current-user-canadministrator-does-not-return-true-in-multisite-if-user-is-a). Could you please replace it with `memberful_is_admin()`? We already use it in `src/syncing.php`.

wordpress/wp-content/plugins/memberful-wp/src/contrib/woocommerce.php
```diff
@@ -18,15 +18,27 @@ public static function instance() {
...
     add_filter( 'woocommerce_add_to_cart_validation', array( $this, 'block_cart_add' ), 30, 3 );
-    add_filter( 'woocommerce_is_purchasable', array( $this, 'not_purchasable'), 20, 2 );
+    add_filter( 'woocommerce_is_purchasable', array( $this, 'is_purchasable'), 20, 2 );
+    add_filter( 'the_content', array( $this, 'remove_erroneous_protection' ), -20 );
+  }
+
+  /**
+   * Removes filtering on the content from other functionalities
+   */
+  function remove_erroneous_protection( $content ) {
```
mrhead: @LiamBailey Why do we do this?

LiamBailey: We do this because when you do add_filter( 'the_content', 'memberful_wp_protect_content', -10 ); the memberful_wp_protect_content catches the_content of all post types. In the case of the WooCommerce integration they want to hide the add to cart button but show the_content which is the product description. So I need to unhook that filter if the post_type is product.

wordpress/wp-content/plugins/memberful-wp/src/contrib/woocommerce.php
LiamBailey: OK I changed it to use that function - https://github.com/memberful/memberful-wp/pull/222/commits/90475736b7181b9de8e514bd00a0995fb96a470a

wordpress/wp-content/plugins/memberful-wp/src/contrib/woocommerce.php
mrhead: Thanks for the explanation!

mrhead:
Thanks!

PR #226, opened 2018-06-15
Set protocol for memberful_sign_in_url based on is_ssl()

mrhead:


mrhead:
Hmm, this might not work correctly for websites behind a load balancer. I'll think about it over the weekend and then decide what to do.

mrhead:
Related: https://core.trac.wordpress.org/ticket/31288

PR #236, opened 2019-03-13
Modified comments section

ChaseWPDEV:
Protect comments feed

### DIFF
--- a/wordpress/wp-content/plugins/memberful-wp/src/comments_protection.php
+++ b/wordpress/wp-content/plugins/memberful-wp/src/comments_protection.php
@@ -29,8 +29,7 @@ function memberful_comments_protection_comments_template() {
  */
 function memberful_user_can_access_comments(){
       // The user most has admin - publisher rights, so we'll not restrict comments access.
-  if ( current_user_can( 'publish_posts' ) )
-    return true;
+  if (current_user_can( 'publish_posts' )) return true;
 
   global $post;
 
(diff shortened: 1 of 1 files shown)


mrhead:
@ChaseWPDEV Thank you for the pull request! 

However it looks that comments feed behaviour isn't consistent with post page:

| Scenario | Comments Feed | Post page |
| --- | --- | --- |
| User is not logged in, post is not protected | Feed is removed ❌ | Comments are shown ✅ |
| User is not logged in, post is protected | Feed is removed ✅ | Comments are not shown ✅ |
| User with access is logged in | Feed is removed ❌ | Comments are shown ✅ |
| User without access is logged in | Feed is removed ✅ | Comments are not shown ✅ |

Basically it looks that feed is never shown even if it should be. @ChaseWPDEV Could you please check it?

ChaseWPDEV:
Quick question regarding the third row @mrhead:
Communications on Codeable were:
 "We're effectively disabling the comments feed for any Memberful-protected posts (because RSS is machine read and the machine won't be signed in). Showing it to signed in users would only be more confusing because their reader couldn't do anything with it. With that in mind, we should probably just return nothing. In other words, we want to disable the comment feed for Memberful protected posts."

Is this no longer the case? Just want to clarify the goal before I make the modifications.

Lauren-Memberful:
@mrhead In case it wasn't clear Patrik, the quote that Chase is referencing above is the instructions that I provided to him via Codeable, and are based on my discussion with @dstrojny.

mrhead:
I'm sorry @ChaseWPDEV. This was a miscommunication on our side.

After reviewing the requirements, the feature really works as intended. I just have one question regarding the code. We should be done here once your answer it.

mrhead:
@ChaseWPDEV Actually I figured out the code question by myself.

However, I've also noticed that `/comments/feed/` is empty if a site has at least one protected post with comments. I'm not sure if this behaviour is desirable. I'll check it internally and I'll come back to you.

mrhead:
@ChaseWPDEV post specific comment feed works as expected 👍. 

However, we need to exclude comments on protected posts from the site comment feed instead of disabling it altogether. Could you please take care of it? Thanks!

ChaseWPDEV:
@mrhead and @Lauren-Memberful Plugin updated. My local tests are showing the following:

/feed: no impact, original plugin behavior intact
/comments/feed: comments from protected posts are omitted, comments from unprotected posts still appear
/IAmAProtectedPost/feed: no feed appears

No variation between a logged in user and non-logged in user, except for /feed where the original plugin behavior is continuing to alter the site's baseline feed as before.

Please let me know if there are any questions or concerns.

Best Regards,
Chase

mrhead:
@ChaseWPDEV Thank you! This looks better, however there is still one edge case.

When a post is protected, but doesn't require any particular plan, then comments are still included in `/comments/feed`:

Screenshot 2019-03-19 at 10 58 59

Otherwise the code looks good except for some formatting inconsistencies.

ChaseWPDEV:
@mrhead and @Lauren-Memberful  Got it :) Modified to meet that edge case.
I also tweaked the code a bit to get closer alignment with the existing code base.

Best Regards,
Chase

mrhead:
Thanks @ChaseWPDEV! It looks that everything is working as expected. I'll let Lauren know that we are done here.

PR #241, opened 2019-06-05
Use BIGINT(20) for IDs and timestamps

mrhead:
[WordPress uses BIGINT(20) for all IDs][1], therefore we should use it in our foreign keys, too. Also, I've decided to update `member_id` and [`last_sync_at`][2], too, to make our table future proof.

[1]: https://codex.wordpress.org/Database_Description
[2]: https://en.wikipedia.org/wiki/Year_2038_problem


mrhead:
It's worth noting that this will be applied only to new installations and maybe we will need to update DB structure for existing users in the future.

PR #244, opened 2019-09-04
Mark restricted posts in admin for subscriber-only

northerner:
Ensure the "Protected by Memberful" message is displayed in the admin
posts list on posts that are restricted to any member with an active
subscription.

This is in addition to the current logic that displays the notice only
if a post is restricted to subscribers of a specific plan.


northerner:
> Nice catch @northerner!
> 
> Could you please do the same for the "All members (active, inactive, or free)." option and then update the README with this change?

Thanks. Made the addition for all members restrictions now, and bump version to 1.49.1.

PR #245, opened 2019-09-13
Only restrict main content on Elementor pages

northerner:
This fix ensures the restricted content filter is only removing content
in the main page or post body when running with Elementor themes.

https://github.com/elementor/elementor/issues/8930

### DIFF
--- a/wordpress/wp-content/plugins/memberful-wp/src/content_filter.php
+++ b/wordpress/wp-content/plugins/memberful-wp/src/content_filter.php
@@ -14,7 +14,7 @@ function memberful_wp_protect_content( $content ) {
     remove_action( "the_content", "FLBuilder::render_content" );
 
     // Remove Elementor action hook
-    remove_action("elementor/frontent/the_content", "memberful_wp_protect_content");
+    remove_action("elementor/frontend/the_content", "memberful_wp_protect_content");
 
     // Remove media enclosures from the RSS feed
     add_filter("rss_enclosure", "__return_empty_string");
--- a/wordpress/wp-content/plugins/memberful-wp/src/content_filter.php
+++ b/wordpress/wp-content/plugins/memberful-wp/src/content_filter.php
@@ -13,6 +13,9 @@ function memberful_wp_protect_content( $content ) {
     // Disable Beaver Builder
     remove_action( "the_content", "FLBuilder::render_content" );
 
+    // Remove Elementor action hook
+    remove_action("elementor/frontent/the_content", "memberful_wp_protect_content");
+
     // Remove media enclosures from the RSS feed
     add_filter("rss_enclosure", "__return_empty_string");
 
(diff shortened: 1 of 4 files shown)


wordpress/wp-content/plugins/memberful-wp/src/content_filter.php
```diff
@@ -13,6 +13,9 @@ function memberful_wp_protect_content( $content ) {
     // Disable Beaver Builder
     remove_action( "the_content", "FLBuilder::render_content" );
 
+    // Remove Elementor action hook
+    remove_action("elementor/frontent/the_content", "memberful_wp_protect_content");
```
mrhead: Is this a typo?

```suggestion
    remove_action("elementor/frontend/the_content", "memberful_wp_protect_content");
```

mrhead: I'm not sure if I'm following here, but is this even needed if it worked without it (e.g. with a typo)?

northerner: Thanks, yeah that was a typo. I didn't manage to reproduce the issue myself, but apparently without removing the action there is a risk of an infinite loop: https://github.com/elementor/elementor/issues/8930#issuecomment-525884269

wordpress/wp-content/plugins/memberful-wp/src/contrib/elementor.php
```diff
@@ -1,3 +1,8 @@
 <?php
 
-add_action('elementor/frontend/the_content', 'memberful_wp_protect_content');
+add_filter('elementor/frontend/builder_content_data', function($data, $post_id) {
+  if (get_queried_object_id() === $post_id) {
```
mrhead: I think that there is an edge case because `get_queried_object_id()` can return the same ID as `$post_id` even if it's not a post. There is a small chance of it happening, but do we want to ignore it?

wordpress/wp-content/plugins/memberful-wp/src/content_filter.php
```diff
@@ -13,6 +13,9 @@ function memberful_wp_protect_content( $content ) {
     // Disable Beaver Builder
     remove_action( "the_content", "FLBuilder::render_content" );
 
+    // Remove Elementor action hook
+    remove_action("elementor/frontend/the_content", "memberful_wp_protect_content");
```
mrhead: @northerner OK, I had to check Elementor's source code, but it looks that this is OK (it's still awful callback hell, but what can we do 🤷‍♂️ 😅).

wordpress/wp-content/plugins/memberful-wp/src/contrib/elementor.php
mrhead: It looks that `get_queried_object_id()` uses `get_queried_object()` which might return one of the following: category, tag, taxonomy, posts page, single post, page, or author so in theory there might be a collision (well, I'm not sure if `builder_content_data` is called for anything else then a post).  Also, the same technique is used directly by Elementor so it's probably OK.

PR #249, opened 2019-12-10
Allow category filtering on the private feeds

northerner:
Add ability to filter by category on the private member feed, using a
new parameter that can be used with the private feed shortcode.

For example:
[memberful_private_rss_feed_link category="news"]

No category parameter preserves current behaviour of listing all posts
the users has access to.

### DIFF
--- a/wordpress/wp-content/plugins/memberful-wp/src/private_user_feed.php
+++ b/wordpress/wp-content/plugins/memberful-wp/src/private_user_feed.php
@@ -102,7 +102,7 @@ function memberful_private_rss_feed_link($success_message = '', $error_message =
   $link = (get_home_url() . '/' . memberful_private_user_feed_get_url_identifier($feedToken) );
 
   if($category)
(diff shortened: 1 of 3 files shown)


wordpress/wp-content/plugins/memberful-wp/src/private_user_feed.php
```diff
@@ -104,6 +101,9 @@ function memberful_private_rss_feed_link($success_message = '', $error_message =
 
   $link = (get_home_url() . '/' . memberful_private_user_feed_get_url_identifier($feedToken) );
 
+  if($category)
+    $link .= '&category=' . $category;
```
mrhead: I would like to suggest to use [`add_query_arg`](https://developer.wordpress.org/reference/functions/add_query_arg/) for this. It probably doesn't matter in this case, but it is a WordPress convention for adding parameters to URLs.

PR #251, opened 2020-03-24
Add block tags to feed

lmansur:
* 04ea560 **Add itunes/gpodcasts block tags to private feed**
  
  By default, iTunes and Google Podcasts index all feeds they encounter.
  Since we offer private feeds, this is not intended behavior. As a way to
  prevent this, we can add some tags that forbid feed indexing.
  
  For iTunes, we should add the <itunes:block> [tag](https://help.apple.com/itc/podcasts_connect/?lang=en#/itcb54353390).
  For Google Podcasts, we should add the <googleplay:block> [tag](https://developers.google.com/search/docs/guides/podcast-management#remove).

* 86fd2f4  **Add opt-in setting to include block tags in feed**
  
  Instead of adding the block tags (04ea560f7b1338f03e5b9b43e1810e61ba7ea8c0) by default,
  we'll use an opt-in setting in the plugin page to control that.

Related to: https://3.basecamp.com/3293071/buckets/5610677/todos/1592569560


lmansur:
image

PR #254, opened 2020-05-01
Add Podcasts shortcode

lmansur:
* 3bc4ca4 **Add shortcode for Podcasts link**
  
  This shortcode will render a link to the Podcasts page in the main app.
  We are also renaming the RSS Feed label to make it extra clear that it
  concerns the Wordpress feed and not the new Podcast feature.


lmansur:
I'll merge this once we remove the feature flag in the app

PR #255, opened 2020-05-07
Sync Podcasts

lmansur:
* 47473d2 **Sync Podcasts with MemberAccount**
  
  The main app now returns the member's feeds in the MemberAccount
  payload. We'll map it to a new column in the user's table.
  
  This will be used later to display the tokenized URL for a given Podcast

* 129934c **Add function to check if user has podcast(s)**
  
  This allows the admin to check if the current user has access to the
  given podcast (or podcasts, if an array is given).

* 13095fc **Sync Podcasts with the app**
  
  The plugin will now listen to podcast Webhooks sent from the main app.
  It will then fetch all the Podcasts from the API and persist them in the
  database.
  
  We'll also sync the Podcasts during plugin activation and
  manual-sync.

* 36138cc **Add a shortcode to return the podcast URL**
  
  This shortcode will return the tokenized feed URL for the current user.
  If the user doesn't have access to the podcast, the shortcode doesn't
  return anything.

### DIFF
--- a/wordpress/wp-content/plugins/memberful-wp/js/memberful-wp-tinymce-plugin.js
+++ b/wordpress/wp-content/plugins/memberful-wp/js/memberful-wp-tinymce-plugin.js
@@ -151,23 +151,23 @@
     );
   }
 
-  function insertPodcastUrl(editor) {
+  function insertFeedUrl(editor) {
     function handleDialogSubmit(editor, id) {
       editor.insertContent(
         "[memberful_podcast_url podcast='"+id+"']"
       );
     }
 
-    function podcastOptions(podcast) {
-      return {text: podcast.name, value: podcast.id};
+    function feedOptions(feed) {
+      return {text: feed.name, value: feed.id};
     };
 
-    var podcasts = window.MemberfulData.podcasts;
-    var podcastList = {
+    var feeds = window.MemberfulData.feeds;
+    var feedList = {
       name: "item",
       type: "listbox",
       label: "Podcast",
-      values: podcasts.map(podcastOptions)
+      values: feeds.map(feedOptions)
     };
 
     editor.windowManager.open({
@@ -175,7 +175,7 @@
       width: 350,
       height: 60,
       body: [
-        podcastList,
+        feedList,
       ],
       onSubmit: function(e) {
         handleDialogSubmit(editor, e.data.item);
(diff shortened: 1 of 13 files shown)


lmansur:
That's a good point, @jcbpl . Fixed in bd66adb

mrhead:
@lmansur I might be a bit confused, but I have a couple of questions.

**TineMCE editor**

It looks that this pull request should update our Memberful in the TinyMCE editor but I don't see any podcasts link in it:

Screenshot 2020-05-12 at 16 01 43

These is a podcast in `window.MemberfulData.feeds` so I presume that it should work.

Screenshot 2020-05-12 at 16 08 35

Btw. I'm using the `podcasts-wp` branch from the Memberful repository.

**Settings page**

Could we add info about podcasts to the settings page?

Screenshot 2020-05-12 at 16 02 31

**Restrict access tool**

I'm not sure if you plan to add this in another pull request, but could (should?) we add podcasts also to the restrict access tool?

Screenshot 2020-05-12 at 16 11 33

lmansur:
@mrhead 
**TinyMCE**: That's odd. It's showing up for me. Are both branches up-to-date? Do you see anything in the Debug logs?

**Settings Page**: Yeah, good idea. Added in f937f9f

**Restrict Access**: We haven't discussed this yet, but it's a possibility that crossed my mind. I think we can address this in a later PR.

okdan:
@lmansur I'm having the same issue as @mrhead - I don't seem to see any of the shortcodes in the interface:
Screen Shot 2020-05-13 at 6 01 07 PM

This is true on branches #254 and #255. Does it have something to do with the version number? Locally it says I'm on 1.53, but that doesn't seem to match what's in the PR. Not sure, as I don't work on this much.

okdan:
I'm having a separate issue where I can't get the plugin to sync the podcasts tothe plugin - so the individual feed short code doesn't display. When I try to run a manual sync in the plugin, it throws an error. My connected app has 1 podcast in it.

Screen Shot 2020-05-13 at 6 15 24 PM

lmansur:
@mrhead I tried resetting my setup, but I'm still not able to reproduce the TinyMCE issue. Do you see anything strange in the logs?

mrhead:
@lmansur I'm sorry, I've missed your previous comment.

I haven't seen anything in the logs before. But if it's working for you then feel free to merge this in.

lmansur:
Deployed the plugin to staging and everything looked good. Merging it.

PR #258, opened 2020-07-01
Add price to the memberful_buy_subscription_link shortcode

mrhead:
This includes the "Link to checkout" TinyMCE dialog for creating
shortcodes. The price field is shown for all plans at the moment. It
looks that it is possible to show it only for "Choose what you pay"
plans, however this change looks order of magnitude more complex and it
might not be worth doing it.

Screenshot 2020-07-01 at 10 38 48

The tooltip is shown only on hover.

https://3.basecamp.com/3293071/buckets/17429243/todolists/2740201364

### DIFF
--- a/wordpress/wp-content/plugins/memberful-wp/js/memberful-wp-tinymce-plugin.js
+++ b/wordpress/wp-content/plugins/memberful-wp/js/memberful-wp-tinymce-plugin.js
@@ -96,7 +96,7 @@
         name: "price",
         type: "textbox",
         label: "Price (optional)",
-        tooltip: "Works only with plans where members are allowed to set price."
+        tooltip: "Works only with plans where members can choose what they pay."
       };
 
       body.push(priceCtrl);
--- a/wordpress/wp-content/plugins/memberful-wp/js/memberful-wp-tinymce-plugin.js
+++ b/wordpress/wp-content/plugins/memberful-wp/js/memberful-wp-tinymce-plugin.js
@@ -72,43 +72,53 @@
   }
 
   function insertCheckoutLinkDialog(editor, options) {
-    var checkoutItemCtrl = {}, linkTextCtrl = {};
-
     function optionsForPurchasables(current) {
       return {text: current.name, value: current.slug};
     };
 
-    options = options || {};
-
-    checkoutItemCtrl = {
+    var checkoutItemCtrl = {
       name: "item",
       type: "listbox",
       label: options.label,
       values: options.choices.map(optionsForPurchasables)
     };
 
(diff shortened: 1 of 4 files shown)


wordpress/wp-content/plugins/memberful-wp/js/memberful-wp-tinymce-plugin.js
```diff
@@ -72,43 +72,53 @@
...
     };
 
+    var body = [checkoutItemCtrl, linkTextCtrl];
+
+    if(options.showPrice) {
+      var priceCtrl = {
+        name: "price",
+        type: "textbox",
+        label: "Price (optional)",
+        tooltip: "Works only with plans where members are allowed to set price."
```
lmansur, line 99: What do you think of using the "choose what you pay" terminology here? Admins might not associate "allowed to set price" with CWYP.

Maybe something like _"Works only with plans where members can choose what they pay"_?

mrhead: I like that! I changed it in b5cede474ddb166278b770c9225a951ffe01180c.

@okdan Are you OK with the copy suggested by Lucas?

okdan: That works for me 👍

okdan:
@mrhead Based on the screenshot, I'm a little concerned about the tooltip on the form field. That's a bit of an unconventional pattern, and I'm concerned it might interfere too much. But I'll pull this down and test it locally to see.

I think the preference would definitely be to only show this field for plans that it applies to. How much extra work would that be?

mrhead:
> Based on the screenshot, I'm a little concerned about the tooltip on the form field. That's a bit of an unconventional pattern,...

It is a standard TinyMCE tooltip and it is shown only on hover.

> I think the preference would definitely be to only show this field for plans that it applies to. How much extra work would that be?

As I said, at least order of magnitude more work. There is a bunch of work needed to support this:

- Add info about CWYP plans to the API.
- Store this information in WordPress.
- Configure the TinyMCE dialog to show/hide the field based on the selected plan.

To be honest, the last point seems the most problematic to me. I played with it yesterday, and while TinyMCE has some API for showing/hiding fields, it was acting pretty weird. I.e. it hid the field, but not the label (???). Then when I tried to hide it manually the dialog size stayed the same and there was just blank space instead of the field.

I could definitely try to implement it, but after playing with it yesterday I don't think that it is worth it. Do you want to try it anyway?

PR #259, opened 2020-07-31
Allow specifying a feed in the feeds_url shortcode

lmansur:
* fdcfa29 **Allow specifying a feed in the feeds_url shortcode**
  
  Now that Memberful supports pre-selecting a Feed using a query string,
  it makes sense for the plugin to support this as well.


lmansur:
Related to https://github.com/memberful/memberful/pull/4925

PR #260, opened 2020-08-05
Fix layout for the Restrict Access tool

mrhead:
Fixes: https://3.basecamp.com/3293071/buckets/9856127/todos/1573668843

Before:

After:


okdan:
Nice - I've been avoiding this to-do for a year! Not sure this matters, but @lmansur just pushed an update to the WP plugin, so make sure it's rebased (or however that works).

PR #261, opened 2020-08-25
Fix bulk restrict category select

northerner:
The JS to dynamically display form elements based on other input values
was not switching based on the element's property value, but was using
the element's attribute value set on page creation.

This would mean that when a select field was used as the dynamic element
for "data-depends-on" it would always return the value it was originally
set to for comparison, and not whatever it was updated to.

https://3.basecamp.com/3293071/buckets/5610905/todos/2960092719

### DIFF
--- a/wordpress/wp-content/plugins/memberful-wp/js/admin.js
+++ b/wordpress/wp-content/plugins/memberful-wp/js/admin.js
@@ -21,7 +21,7 @@ jQuery(document).ready(function($){
 
     var callback = function() {
       var self = $(this);
-      var field_value = self.attr('value');
+      var field_value = self.val();
       var dependent_can_be_shown = false;
       var fieldType = self.prop('tagName') == 'SELECT' ? 'select' : self.prop('type');
       var checked = self.is(':checked');


lmansur:
I was able to reproduce this on Jen's Bakery, but not locally. Were you able to?

northerner:
> I was able to reproduce this on Jen's Bakery, but not locally. Were you able to?

Yeah, without this fix the categories won't show for me. That was in Chrome 84.

Does the existing category select work for you in the dev environment?

lmansur:
@northerner yep!

The categories were hidden on page load almost instantly, so I guess it has something to do with how fast the JS is loaded locally vs. remotely?

northerner:
`The categories were hidden on page load almost instantly, so I guess it has something to do with how fast the JS is loaded locally vs. remotely?`

And they display fine when you select the "All posts from a category" option in the drop down?

lmansur:
Yes:
wp-categories-bulk

northerner:
@mrhead, Does this issue happen for you on Chrome too?

It is not a bug for me in Firefox or Safari, but in Chrome this bug happens on Jen's Bakery and locally.

The fact that I'm getting it, Lauren sees it, and so are our customers, makes me think we need this fix.

mrhead:
@northerner I can reproduce this locally and on Jen's Bakery with Firefox 80. Also, this PR fixes it for me.

I'm attaching some GIFs, just to be sure that we are talking about the same thing.

Firefox + current master:

Screen Recording 2020-08-27 at 13 59 28

Firefox + this PR:

Screen Recording 2020-08-27 at 14 00 07

p.s. sorry for the GIF quality, the mov -> gif conversion slowed them down.

PR #268, opened 2020-12-16
Make private user feed title configurable

mrhead:
This is a continuation of #264. I've introduced a new helper function
`memberful_get_bloginfo_rss()` which makes getting `bloginfo_rss`
attributes easier.

https://3.basecamp.com/3293071/buckets/9856127/messages/3301743790


lmansur (approved):
Nice 👍

Issue #223, opened 2018-02-23
Can't select first on "Link to download" dropdown list


dstrojny:
I wanted to **embed a Download** on a page on my WordPress site. I tried to select the first item from the dropdown list, but was unable to do so. I can only select the first item if I choose another Download first and then return to the first item.

Tested with Chrome and Safari.

![link-to-download](https://user-images.githubusercontent.com/34775566/36500151-7966a5fe-1711-11e8-9121-4b8c496b7d47.gif)

PR #269, opened 2021-01-11
Fix link to download select

northerner:
Using the "text" property as a placeholder for the download select was
breaking the selection of the first item in the list. This is a labeled
select box so a placeholder value was not necessary anyway.

#223


mrhead:
@northerner What do we show instead of "Choose download"? TinyMCE API is hard to follow. 😳

northerner:
> @northerner What do we show instead of "Choose download"? TinyMCE API is hard to follow. 😳

It just defaults to the first download available, [like podcasts dropdown](https://github.com/memberful/memberful-wp/blob/6397e2ba866ce212c861e1141e9d0e60ba9412b3/wordpress/wp-content/plugins/memberful-wp/js/memberful-wp-tinymce-plugin.js#L69).

mrhead (approved):
👍

PR #270, opened 2021-01-12
Add new option to hide the admin toolbar

northerner:
When enabled the admin toolbar will only be displayed for users with
post editing permissions. All other users, including those created by
Memberful, will not see the toolbar.

The option is enabled by default.

toolbar_options


lmansur (approved):
👍

Might be worth removing the 3rd-party plugin from the [docs](https://memberful.com/help/integrate/services/wordpress/other-plugins-recommended/#install-hide-admin-toolbar-plugin).

PR #272, opened 2021-01-15
Filter account menu items based on signed-in state

northerner:
Adding an option (on by default) to filter out "Sign in" links from
menus when already signed in, and filter out "Sign out" links when
signed out.


mrhead (approved):
Nice work!

PR #273, opened 2021-01-15
Add docker compose setup

northerner:
Beginning alternative setup for development to remove dependencies on
VCCW and Vagrant.


mrhead:
Can't wait for this! 👍

docker-compose.yml
```diff
@@ -28,6 +28,8 @@ services:
        - "./wordpress:/var/www/html"
        - ".:/root"
        - "./site.conf:/etc/apache2/sites-available/000-default.conf"
+     extra_hosts:
+       - "apps.memberful.localhost:host-gateway"
```
northerner, line 32: Running `docker exec memberful-wp_wordpress_1 cat /etc/hosts` after building with this option shows this added in the hosts file:

`192.168.65.1	apps.memberful.localhost`

But trying to curl that domain results in nothing. For comparison, a curl to `host.docker.internal` from inside the container gives me the "unknown app" response I would expect from post 80 of my host machine.

README.md
```diff
@@ -2,9 +2,11 @@
 
 ## Development environment
 
-This project uses [VCCW](http://vccw.cc/). 
+This project uses [Docker](https://www.docker.com/) for development.
 
-To get up and running install [Vagrant](http://vagrantup.com) and [Virtualbox](http://virtualbox.org), then run `vagrant up` in the plugin's root directory.
+To get up and running install [Docker](https://www.docker.com/get-started), then run `docker-compose up -d` in the plugin's root directory.
+
+For the inital Wordpress setup run `chmod +x docker-provision.sh` and `./docker-provision.sh`.
```
lmansur, line 9: Could we add the proper permissions through Git?

docker-compose.yml
```diff
@@ -0,0 +1,36 @@
...
+       - db
+     image: wordpress:latest
+     ports:
+       - "8181:80"
+     restart: always
+     environment:
+       WORDPRESS_DB_HOST: db:3306
+       WORDPRESS_DB_USER: wordpress
+       WORDPRESS_DB_PASSWORD: wordpress
+       WORDPRESS_DB_NAME: wordpress
```
lmansur, line 26: Can we re-use the `envfile` on the root directory here? It should make it easier to add other env vars such as `WORDPRESS_DEBUG`

lmansur (approved):
Nice work, thanks for handling this!

README.md
northerner: Yeah, good call!

PR #275, opened 2021-01-26
Add content access by term (category or tag)

northerner:
Restricting access to posts automatically by category or tag. With
related marketing content set at the term level.

Post restrictions take precedence, if any terms with restrictions are
added to a restricted post the post-specific restrictions will be used
and term restrictions ignored.

### DIFF
--- a/wordpress/wp-content/plugins/memberful-wp/src/acl.php
+++ b/wordpress/wp-content/plugins/memberful-wp/src/acl.php
@@ -1,6 +1,7 @@
 <?php
 require_once MEMBERFUL_DIR . '/src/acl/helpers.php';
 require_once MEMBERFUL_DIR . '/src/acl/post_options.php';
+require_once MEMBERFUL_DIR . '/src/acl/term_options.php';
 
 /**
  * Determines the set of post IDs that the current user cannot access
@@ -13,6 +14,16 @@ require_once MEMBERFUL_DIR . '/src/acl/post_options.php';
  * @return array Map of post ID => post ID
  */
 function memberful_wp_user_disallowed_post_ids( $user_id ) {
+  $acl = get_option( 'memberful_acl', array() );
+  return memberful_wp_user_disallowed_ids_from_acl( $user_id, $acl);
+}
+
+function memberful_wp_user_disallowed_term_ids( $user_id ) {
+  $acl = get_option( 'memberful_term_acl', array() );
+  return memberful_wp_user_disallowed_ids_from_acl( $user_id, $acl);
+}
+
+function memberful_wp_user_disallowed_ids_from_acl( $user_id, $acl ) {
   static $ids = array();
 
   $user_id        = (int) $user_id;
@@ -21,7 +32,6 @@ function memberful_wp_user_disallowed_post_ids( $user_id ) {
   if ( isset( $ids[$user_id] ) )
     return $ids[$user_id];
 
-  $acl = get_option( 'memberful_acl', array() );
   $global_product_acl = isset( $acl['product'] ) ? $acl['product'] : array();
   $global_subscription_acl = isset( $acl['subscription'] ) ? $acl['subscription'] : array();
   $posts_for_any_registered_users = memberful_wp_get_all_posts_available_to_any_registered_user();
@@ -195,7 +205,29 @@ function memberful_wp_extract_slug_ids_and_user($args) {
  * @param integer $post_id ID of the post that should have access checked
  */
 function memberful_can_user_access_post( $user, $post ) {
-  $restricted_posts = memberful_wp_user_disallowed_post_ids( $user );
+  if ( !empty( memberful_terms_restricting_post( $user, $post )))
+    return false;
 
+  $restricted_posts = memberful_wp_user_disallowed_post_ids( $user );
   return ! isset( $restricted_posts[$post] );
+}
+
+function memberful_terms_restricting_post( $user, $post ) {
+  $restricted_terms = array_values( memberful_wp_user_disallowed_term_ids( $user ));
+  return array_intersect( $restricted_terms, memberful_wp_get_category_and_tag_ids_for_post( $post ));
+}
+
+function memberful_wp_get_category_and_tag_ids_for_post( $post ) {
+  $categories = get_the_terms( $post, "category" );
+  $tags = get_the_terms( $post, "post_tag" );
+
+  $categories = $categories ? $categories : array();
+  $tags = $tags ? $tags : array();
+  $terms = array_merge( $categories, $tags );
+
+  if ( !empty( $terms )) {
+    $terms = wp_list_pluck( $terms, "term_id" );
+  }
+
+  return $terms;
 }
--- a/wordpress/wp-content/plugins/memberful-wp/src/acl/term_options.php
+++ b/wordpress/wp-content/plugins/memberful-wp/src/acl/term_options.php
@@ -1,0 +1,45 @@
+<?php
+
+function memberful_wp_get_term_available_to_any_registered_users( $term_id ) {
+  return get_term_meta( $term_id, 'memberful_available_to_any_registered_user', TRUE ) === "1";
+}
+
+function memberful_wp_set_term_available_to_any_registered_users( $term_id, $is_viewable_by_any_registered_users ) {
+  update_term_meta( $term_id, 'memberful_available_to_any_registered_user', $is_viewable_by_any_registered_users);
+
+  $globally_viewable_by_any_registered_users = memberful_wp_get_all_terms_available_to_any_registered_user();
+
+  if ( $is_viewable_by_any_registered_users ) {
+    $globally_viewable_by_any_registered_users[$term_id] = $term_id;
+  } else {
+    unset($globally_viewable_by_any_registered_users[$term_id]);
+  }
+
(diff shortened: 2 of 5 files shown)


wordpress/wp-content/plugins/memberful-wp/src/acl/term_options.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+function memberful_wp_get_term_available_to_any_registered_users( $term_id ) {
+  return get_term_meta( $term_id, 'memberful_available_to_any_registered_user', TRUE ) === "1";
+}
```
lmansur, line 5: I think the function's name is a bit misleading as it doesn't get a term available to any user but instead checks if the given term is available to any user. Not sure if there's a convention for this in PHP land, but maybe `memberful_wp_is_term_available_to_any_registered_users`?

Now that I thought about it some more, I think I understand the intention: to get the term's column `available_to_any_registered_user`. But I think it's a bit ambiguous, imo.

wordpress/wp-content/plugins/memberful-wp/src/acl/term_options.php
```diff
@@ -0,0 +1,45 @@
...
+  update_option( 'memberful_terms_available_to_any_registered_user', $globally_viewable_by_any_registered_users );
+}
+
+function memberful_wp_get_all_terms_available_to_any_registered_user() {
+  return get_option( 'memberful_terms_available_to_any_registered_user', array() );
+}
+
+function memberful_wp_get_term_available_to_anybody_subscribed_to_a_plan( $term_id ) {
+  return get_term_meta( $term_id, 'memberful_available_to_anybody_subscribed_to_a_plan', TRUE ) === "1";
+}
```
lmansur, line 27: Similar to the comment above about the function's name

mrhead (commented):
I had a question regarding memoization. Otherwise this looks good to me given you manually tested it.

wordpress/wp-content/plugins/memberful-wp/src/acl.php
```diff
@@ -13,15 +14,33 @@
  * @return array Map of post ID => post ID
  */
 function memberful_wp_user_disallowed_post_ids( $user_id ) {
-  static $ids = array();
+  if ( isset( $disallowed_post_ids )) {
+    return $disallowed_post_ids;
+  } else {
+    static $disallowed_post_ids = array();
```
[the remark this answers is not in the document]
northerner: Yeah, effectively cached for this request.

lmansur:
@northerner I noticed an issue of blank posts, but still trying to come up with the reproduction steps

lmansur (approved):
🎉

wordpress/wp-content/plugins/memberful-wp/src/acl.php
```diff
@@ -195,7 +214,54 @@ function memberful_wp_extract_slug_ids_and_user($args) {
  * @param integer $post_id ID of the post that should have access checked
  */
 function memberful_can_user_access_post( $user, $post ) {
-  $restricted_posts = memberful_wp_user_disallowed_post_ids( $user );
+  if ( !empty( memberful_terms_restricting_post( $user, $post ))) {
```
mrhead, line 217: @northerner I just started digging into the ACL code, but I have a feeling that we shouldn't modify `memberful_can_user_access_post()`. Instead of that we should add term restrictions checks into `memberful_wp_user_disallowed_ids_from_acl()` which is used by `memberful_wp_user_disallowed_post_ids()`.

That should work nicely with merging restrictions settings from posts and terms as we agreed on it in Basecamp.

wordpress/wp-content/plugins/memberful-wp/src/metabox.php
```diff
@@ -257,3 +321,41 @@ protected function meta_field() {
     return 'memberful_'.$this->_entity;
   }
 }
+
+class Memberful_Term_ACL extends Memberful_Post_ACL {
```
mrhead, line 325: @northerner I have a feeling that `Memberful_Term_ACL` shouldn't extend `Memberful_Post_ACL`. Instead of that we should probably create a base ACL class and both, `Memberful_Post_ACL` and `Memberful_Term_ACL` should extend it.

However, this can wait until we solve the restriction issues.

wordpress/wp-content/plugins/memberful-wp/src/acl.php
```diff
@@ -195,7 +214,135 @@ function memberful_wp_extract_slug_ids_and_user($args) {
  * @param integer $post_id ID of the post that should have access checked
  */
 function memberful_can_user_access_post( $user, $post ) {
```
mrhead, line 216:
```suggestion
function memberful_can_user_access_post( $user_id, $post_id ) {
```

It looks that both arguments are IDs so we should rename that. Otherwise `"user-{$user}--post-{$post}"` from below gets pretty confusing. At first I thought that we are using objects in cache keys.

wordpress/wp-content/plugins/memberful-wp/src/acl.php
```diff
@@ -195,7 +214,135 @@ function memberful_wp_extract_slug_ids_and_user($args) {
...
-  $restricted_posts = memberful_wp_user_disallowed_post_ids( $user );
+  $cache_result = wp_cache_get( "user-{$user}--post-{$post}", "memberful_post_access", $found = false );
```
mrhead, line 217: Is there a performance benefit to this change? The benchmarks showed that post protection with terms wasn't slower than the original implementation. Also, this adds additional complexity and we need to take care of cache invalidation which I don't see anywhere. Is it present?

Anyways, I would vote against adding caching in this PR. It's pretty complex even without caching so if we want to introduce caching for checking permission, then we should probably do it in a separate pull request.

mrhead: Also, I don't think that this actually works as expected. `$found` should be passed as a reference, however, it should be the fourth argument not third. I'm pretty sure that here we just set `$found = false` and pass `false` as the third argument with the fourth argument being empty.

wordpress/wp-content/plugins/memberful-wp/src/marketing_content.php
```diff
@@ -4,14 +4,31 @@
...
-  return apply_filters( 'memberful_marketing_content', $memberful_marketing_content );
+  $user_id = is_user_logged_in() ? get_current_user_id() : 0;
+  $restricted_posts = memberful_wp_user_disallowed_post_ids( $user_id );
+
+  if ( isset( $restricted_posts[$post_id] )) {
+    $memberful_marketing_content = get_post_meta( $post_id, MEMBERFUL_MARKETING_META_KEY, TRUE );
+    return apply_filters( 'memberful_marketing_content', $memberful_marketing_content );
+  } else {
+    $terms = memberful_terms_restricting_post( $user_id, $post_id );
+    return memberful_term_marketing_content( reset( $terms ));
```
mrhead, line 15: I'm not sure if we should call `reset()` here. If it is really needed, then we should call it in `memberful_term_marketing_content()` before we return the array.

mrhead: Ah, never mind. I've misunderstood what this should do.

wordpress/wp-content/plugins/memberful-wp/src/metabox.php
mrhead: @northerner Do you have any thoughts on this? I feel strongly that we should refactor this before we merge this PR in.

mrhead: I'm working on a PR for this. Will share it soon.

mrhead: https://github.com/memberful/memberful-wp/pull/298

wordpress/wp-content/plugins/memberful-wp/src/acl.php
mrhead: This is actually not related to this PR so we can do it later.

wordpress/wp-content/plugins/memberful-wp/src/acl.php
```diff
@@ -195,7 +214,116 @@ function memberful_wp_extract_slug_ids_and_user($args) {
...
+}
+
+function memberful_wp_post_viewable_by_any_subscriber( $post, $terms_for_post ) {
+  $posts_for_anybody_subscribed_to_a_plan = memberful_wp_get_all_posts_available_to_anybody_subscribed_to_a_plan();
+  $terms_for_anybody_subscribed_to_a_plan = memberful_wp_get_all_terms_available_to_anybody_subscribed_to_a_plan();
+
+  return ( in_array( $post, $posts_for_anybody_subscribed_to_a_plan ) || array_intersect( $terms_for_post, $terms_for_anybody_subscribed_to_a_plan ));
+}
+
+function memberful_terms_restricting_post( $user, $post ) {
```
mrhead, line 287: @northerner If I'm correct, we use this function to find a term for displaying marketing content. We never need all the terms restricting a post, only the first one. Therefore we could maybe rename this function and return from it as soon as we find a term so we don't do unnecessary work.

This is not blocking at this point, but I'll try to create a PR with this change.

mrhead: https://github.com/memberful/memberful-wp/pull/299

PR #276, opened 2021-01-29
Allow ajax admin access when dashboard blocked

northerner:
This prevents conflicts with plugins that use the Wordpress ajax admin
endpoint, but still allows us to redirect members from visiting the
Wordpress dashboard with the new option enabled.


mrhead (approved):
👍

lmansur (approved):

PR #280, opened 2021-02-26
Update filter priority

ChaseWPDEV:
Negative action priorities can often result in unexpected behavior on filtering.

In this case, the -10 priority filter on your 'memberful_wp_protect_content' function applied to the 'the_content' filter is overwritten by elementor reaccessing the post's data as part of the blogroll loop.

More common behavior for membership plugins is to hook into the content at a later entry priority (I used 100 here, which matches Restrict Content Pro), and then ensure key filters (such as 'do_shortcodes') are applied to the marketing content.

### DIFF
--- a/wordpress/wp-content/plugins/memberful-wp/src/content_filter.php
+++ b/wordpress/wp-content/plugins/memberful-wp/src/content_filter.php
@@ -1,6 +1,6 @@
 <?php
 
-add_action( 'the_content', 'memberful_wp_protect_content', -10 );
+add_action( 'the_content', 'memberful_wp_protect_content', 100 );
 
 function memberful_wp_protect_content( $content ) {
   global $post;
@@ -27,3 +27,5 @@ function memberful_wp_protect_content( $content ) {
 
   return $content;
 }
+
+add_filter( 'memberful_wp_protect_content', 'do_shortcode' );


northerner:
Thanks @ChaseWPDEV, this fixes the Elementor Pro issue.

It looks like we've been using the -10 priority for 7 years, the [original commit](https://github.com/memberful/memberful-wp/commit/7f8a97ff5abdb77f285af0cb7bc6f98ef85bcb09) that introduced it suggested it might be needed as we add the marketing content before doing the filtering. Have there been Wordpress changes that would mean this priority can change now?

ChaseWPDEV:
Good question @northerner . There hasn't been a significant change specific to this filter in WP Core, persay--but a rise in page builders does change some of the considerations of how you select your filter priority.

**-An early filter** (like the -10 that you previously had) ensures that your content gets injected very early in the filtering process. Which allows it to then be subsequently processed by all of the other filters the WordPress will run on it. It's good--in that your marketing content is then subject to all the same filters that other content is. It's bad in that your marketing content can then get overwritten by subsequent filters.

**-A later filter** results in your marketing content not being subjected to all the same filters the rest of your content is--meaning that marketing content built with thrid party page builders might need to have those filters re-applied to render correctly. The benefit of a later filter ensures that your filter provides the authoritative source of the content (the marketing content)--which is typically the behavior we expect from "guard" filters.

**What was happening:** Elementor was overwriting the content filters in the blog loop. I suspect their filter was intentionally replacing the content to ensure that excerpt lengths weren't chopping their markup up and breaking it--but by doing so, they're breaking the way filters are supposed to cascade. IMO, they're the ones that are wrong here--but they're a bigger fish, and I do understand why they made the decision that they did.

**Is this a backward-compatibility breaking change?** Yes. Depending on the way your customers were generating marketing content and interacting with the memberful plugin's filters (ie custom code), this definitely has the potential to cause customers to need to rework their marketing content.

**Could we do something else instead?** Yes. This fix is very meta, reorienting the overall filtering philosophy for your plugin towards something that will endure similar challenges in the future. Instead, we could do an Elementor specific fix where--when your plugin fires its filter at -10--it removes the Elementor filter that's overwriting the content. This fix would not cause backward compatibility issues but would be much more brittle to failure if Elementor changes its filters even a little bit. (removing filters requires a high degree of specificity).

**Alternatively**, there is a third, middle ground. I found the fix worked at priority filters as low as 11 for your marketing content. Selecting a lower number than 100 would still be backward-compatibility breaking--but slightly less so as the marketing content would be run through a higher percentage of filters.

A lot of considerations there--please take your time to talk it over internally, and let me know if you'd like to pursue one of the different routes for a fix.

northerner:
Great explanation @ChaseWPDEV, thanks for the extra detail. I'll see if the rest of the engineering team have opinions on the fix.

PR #281, opened 2021-03-04
Add uncommitted changes warning to release script

northerner:
The release script does not work from the git main branch as target for
release, it requires you to be on the main branch, but it copies the
plugin folder to a temporary directory for pushing to Wordpress with
SVN.

This warning is to alert anyone running the script that they have
uncommitted changes in the plugin folder.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -29,6 +29,12 @@ check_current_branch() {
   fi
 }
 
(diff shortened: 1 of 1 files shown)


mrhead (approved):
👍

PR #284, opened 2021-03-12
Elementor fix

ChaseWPDEV:


ChaseWPDEV:
Need to add a few more functions--will resubmit.

PR #287, opened 2021-04-14
Enable compatibility with some Gutenberg Blocks

ChaseWPDEV:
Added do_blocks as an early filter to memberful_wp_protect content to enable compatibility with some Butenberg blocks, including gutenberg blocks created by Recipe Card Blocks by WPZOOM


lmansur:
Thanks @ChaseWPDEV !

PR #290, opened 2021-05-20
Remove obsolete provisioning task

mrhead:
This task is not needed anymore because puma-dev takes care of
`.localhost` resolving. Also, the task is broken anyways because it
doesn't add any entries to `/etc/hosts`.


mrhead:
Never-mind. This task is executed on the virtual machine and these entries are in fact needed.

🤦‍♂️🤦‍♂️🤦‍♂️

PR #292, opened 2021-05-20
Do not create wordpress/.gitignore

mrhead:
We already have a `.gitignore` file in the main directory and
`wordpress/.gitignore` clashes with it. As a result, some new files
added to `wordpress/wp-content/themes` are not ignored by Git. This is
pretty annoying because every time I run `vagrant up --provision`
`wordpress/wp-content/themes/index.php` gets added and Git lists it as
an untracked file.

After this is merged in, I suggest that you remove your local
`wordpress/.gitignore` file.


lmansur (approved):
Thank you!

PR #293, opened 2021-05-20
Set redirect_to param on OAuth requests

jcbpl:
This works around Chrome's SameSite cookie restriction when
?memberful_endpoint=auth links open in the overlay.


northerner (approved):

jcbpl:
@lmansur Good point, done in c2ddcf1.

PR #295, opened 2021-05-24
Recursion fix

ChaseWPDEV:
Only allow memberful_wp_content_protect to filter any given content one time, using the doing_filter guard stament.


mrhead:
Thanks Chase!

PR #301, opened 2021-06-01
Add custom domain option

jcbpl:
If enabled, use custom domain to generate HTML-format URLs, and allow
redirects to the hostname.

### DIFF
--- a/wordpress/wp-content/plugins/memberful-wp/src/core-ext.php
+++ b/wordpress/wp-content/plugins/memberful-wp/src/core-ext.php
@@ -21,17 +21,23 @@ function memberful_wp_nonce_field( $action ) {
  * @param array $content A set of websites that can be redirected to
  * @return array The $content plus Memberful domain
  */
-function memberful_wp_allowed_hosts( $content ) {
+function memberful_wp_allowed_hosts( $hosts ) {
   $site = get_option( 'memberful_site' );
+  $custom_domain = get_option( 'memberful_custom_domain' );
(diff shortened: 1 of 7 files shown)


wordpress/wp-content/plugins/memberful-wp/src/core-ext.php
```diff
@@ -21,17 +21,23 @@ function memberful_wp_nonce_field( $action ) {
...
 
-  if ( !empty( $site ) ) {
-    $memberful_url = parse_url( $site );
+  if ( $site ) {
+    $components = parse_url( $site );
 
-    if ( $memberful_url !== false )
-      $content[] = $memberful_url['host'];
+    if ( $components ) {
+      $hosts[] = $components['host'];
```
lmansur, line 32: Non-blocking, but is there a difference between this and `array_push($hosts, $components['host'])`? I find the latter easier to understand.

lmansur (approved):

wordpress/wp-content/plugins/memberful-wp/src/core-ext.php
lmansur: Answering my own question: looks like there isn't a difference: https://stackoverflow.com/questions/676677/how-to-add-elements-to-an-empty-array-in-php

jcbpl: Good call 👍 dd91841

PR #303, opened 2021-06-18
Restrict access by custom taxonomies

mrhead:
Previously we supported only categories and post tags. Since this commit
we will support all public taxonomies visible in menu (we use the same
filter for supporting custom post types).


mrhead:
@lmansur @northerner This is a friendly reminder to review these changes.

Categories and post tags are WordPress core taxonomies. It means that we can use regular taxonomy methods to work with them, and by doing that, we can also support custom taxonomies.

lmansur (approved):
Sorry for the delay! LGTM

mrhead:
Thanks, @lmansur.

northerner:
Sorry @mrhead, missed the notification for this. Feel free to ping me on Basecamp next time if you're waiting on a review.

Have not tested out these changes myself, but the code makes sense to me.

mrhead:
No problem, @northerner. I'll ping on Basecamp next time :).

PR #305, opened 2021-09-06
Non memberful user patch

ChaseWPDEV:
Hi @northerner 

I've got a non-project-related pull request here that I think will help some of our mutual customers.

There are a couple of functions that throw Warnings based on functions returning empty strings instead of arrays when a logged-in user doesn't have a memberful account (e.g. a site admin, or editor). I've added a couple "elvis operators" in this pull request to ensure arrays are returned instead of empty strings.


northerner:
Hi @ChaseWPDEV, thanks for this. Could you rebase it against the latest main branch first please? Would be good to just bring in one commit for the update.

PR #306, opened 2021-09-07
Use MariaDB instead of MySQL

mrhead:
The official Docker image for MySQL doesn't support the ARM architecture
thus it doesn't work with Apple M1 chips.

This change is a drop-in replacement. We could even keep using the old
`MYSQL_*` environment variables, but I changed them to `MARIADB_*` to be
more explicit.

- https://hub.docker.com/_/mariadb/
- https://hub.docker.com/_/mysql/

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,33 +1,33 @@
-version: '3.3'
+version: "3.3"
 
 services:
-   db:
-     image: mysql:5.7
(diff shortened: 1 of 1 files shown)


northerner (approved):
Happy to escape Oracle-land. 👍

lmansur (approved):

PR #309, opened 2021-09-07
Resolve to arrays for non-logged in users

ChaseWPDEV:
@northerner Reworked the previous pull request down to 2 commits :)


northerner:
This has since been fixed in other commits, sorry I missed this one!

PR #311, opened 2021-12-06
Prevent concurrent member mapping

lmansur:
A successful checkout can send up to three webhooks (new member,
subscription, and order), and now that the app sends webhooks in
parallel, it's possible for the plugin to create duplicate users because
of a race condition.

This commit prevents that by using an [explicit lock][1] scoped to the
member ID. The first request to acquire the lock performs the sync while
the others immediately give up.

When the request is made in an OAuth context, we wait up to 10 seconds
for the lock as we need to fetch a `User` instance to sign in.

https://3.basecamp.com/3293071/buckets/24350514/todos/4221739866

[1]: https://dev.mysql.com/doc/refman/5.7/en/locking-functions.html#function_get-lock

### DIFF
--- a/wordpress/wp-content/plugins/memberful-wp/src/authenticator.php
+++ b/wordpress/wp-content/plugins/memberful-wp/src/authenticator.php
@@ -81,7 +81,8 @@ class Memberful_Authenticator {
 
       $account = $this->get_member_data( $tokens->access_token );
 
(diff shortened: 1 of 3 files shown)


mrhead (approved):
Nice work @lmansur!

Does this work with both MyISAM and InnoDB?

lmansur:
@mrhead yep, it works on any storage engine

PR #312, opened 2022-01-03
Update readme with useful dev commands

northerner:
Adding instructions for clearing out the mysql volume and using the
WP-CLI tool.


lmansur (approved):
Nice!

PR #313, opened 2022-01-10
Return empty array if user_meta is empty or false

lmansur:
https://3.basecamp.com/3293071/buckets/5610905/todos/4513433623

### DIFF
--- a/wordpress/wp-content/plugins/memberful-wp/src/acl.php
+++ b/wordpress/wp-content/plugins/memberful-wp/src/acl.php
@@ -114,11 +114,11 @@ function memberful_wp_generate_user_specific_acl_from_global_acl( $users_entitie
  * @return array member's products
  */
 function memberful_wp_user_downloads( $user_id ) {
-  return get_user_meta( $user_id, 'memberful_product', TRUE );
+  return memberful_wp_get_user_meta_for_acl( $user_id, 'memberful_product' );
 }
 
 function memberful_wp_user_feeds($user_id) {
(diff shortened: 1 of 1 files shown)


lmansur:
@mrhead @northerner @jcbpl mind taking a look at this one? I'd like to release this fix today if possible

wordpress/wp-content/plugins/memberful-wp/src/acl.php
```diff
@@ -129,7 +129,21 @@ function memberful_wp_user_feeds($user_id) {
...
+
+/**
+ * `get_user_meta` may return `false` or `""` if the given $user_id
+ * doesn't exist or is invalid. Because the expected ACL interface is an array,
+ * we return an empty one.
+ */
+function memberful_wp_get_user_meta_for_acl($user_id, $meta_key, $single = TRUE) {
+  $meta = get_user_meta($user_id, $meta_key, $single);
+
+  if ($meta == false || $meta == "")
```
mrhead, line 143: According to [PHP's type comparison tables](https://www.php.net/manual/en/types.comparisons.php), `"" == false` is `true` so we can use:

```suggestion
  if ($meta == false)
```

mrhead (approved):
I'm sorry, @lmansur, I've missed a notification about this PR.

It looks good to me. I had one non-blocking comment.

wordpress/wp-content/plugins/memberful-wp/src/acl.php
lmansur: Good to know, thanks!

PR #314, opened 2022-02-01
Filter post access in Private RSS Feed

lmansur:
https://3.basecamp.com/3293071/buckets/5610905/messages/4565347317


mrhead (approved):

northerner (commented):
@lmansur Is the intent here to allow any plugin user to define their own memberful_can_user_access_rss_post filter? The default just allows all posts.

lmansur:
> Is the intent here to allow any plugin user to define their own memberful_can_user_access_rss_post filter? The default just allows all posts.

@northerner yep, that's it. We don't want to change the default but allow plugin users to customize the logic if they want to.

lmansur:
@northerner LMK if you have any other thoughts or if it's good to go

northerner (approved):
Good to go. 👍

PR #318, opened 2022-04-12
3x Feature Implementation

ChaseWPDEV:
Features implemented:
- Auto Sync Display name with Memberful Fullname
- Limit active logged in Sessions per user
- Option for "Global Marketing Content"

028afd9967ec5185fe6ead5cd857a398179a2c72 contains all functional changes

3c55535b2b839572665e390d2546710324ba3a8c contains automated wpcs code formatting changes on modified/added files only.

### DIFF
--- a/wordpress/wp-content/plugins/memberful-wp/memberful-wp.php
+++ b/wordpress/wp-content/plugins/memberful-wp/memberful-wp.php
@@ -3,14 +3,14 @@
 Plugin Name: Memberful WP
 Plugin URI: http://github.com/memberful/memberful-wp
 Description: Sell memberships and restrict access to content with WordPress and Memberful.
-Version: 1.69.0
+Version: 1.69.1
 Author: Memberful
 Author URI: http://memberful.com
 License: GPLv2 or later
  */
 
 if ( ! defined( 'MEMBERFUL_VERSION' ) )
-  define( 'MEMBERFUL_VERSION', '1.69.0' );
+  define( 'MEMBERFUL_VERSION', '1.69.1' );
 
 if ( ! defined( 'MEMBERFUL_PLUGIN_FILE' ) )
   define( 'MEMBERFUL_PLUGIN_FILE', __FILE__ );
--- a/wordpress/wp-content/plugins/memberful-wp/readme.txt
+++ b/wordpress/wp-content/plugins/memberful-wp/readme.txt
@@ -4,7 +4,7 @@ Tags: memberful, member, membership, memberships, recurring payments, recurring 
 Requires at least: 3.6
 Tested up to: 6.0
 Requires PHP: 7.0
-Stable tag: 1.69.0
+Stable tag: 1.69.1
 License: GPLv2 or later
 
 Sell memberships and restrict access to content with WordPress and Memberful.
@@ -48,6 +48,10 @@ Glad you asked! We manage development of the plugin over at the [Memberful WP Gi
 3. Simple sign in and account management widget.
 
 == Changelog ==
+
+= 1.69.1 =
+
+* Fix PHP warnings
 
 = 1.69.0 =
 


mrhead (changes requested):
Hey @ChaseWPDEV,

Thanks for the pull request!

It contains three different features. Since we did not specify it before, let's leave it as it is. 

However, please create a separate pull request for each feature for future work. It makes reviewing code and processing feedback easier.

Also, I'm OK with the automatic formatting of modified files. But this project uses spaces instead of tabs and adding tabs to modified files breaks consistency.

As far as I know, most of the PHP world prefers tabs. However, in the future, please use spaces (unless we migrate the whole repository to tabs, which we don't plan to do).

Now to the features.

### Updating display names

This works great 👍.

### Limiting the number of simultaneous logins

I've set the limit to one session, but I was able to sign in from two different browsers. I was logged out from the first session only after signing in from the third browser.

Could you please check/fix it?

### Global marketing content

`<p>` in `This feature requires <p> tags in your posts to detect which content to use.` is not escaped and it breaks the page layout:

Screenshot 2022-04-14 at 15 31 58

Also, could you add some between the box and the button? Like we do it in other tabs:

Screenshot 2022-04-14 at 15 33 02

Screenshot 2022-04-14 at 15 33 11

Now, regarding the functionality. I tried to test different content scenarios.

In general, the feature works fine. However, if the post doesn't contain two or more paragraphs then things break.

If I include just one paragraph (saying "First paragraph"), then the content is incorrectly parsed and I see this code:

```html
<div class='[memberful-global-teaser-content]()'>
<p>F</div>
```

Or if I start a post with a video embed and don't add two more paragraph then I see this code:

```html
<div class='[memberful-global-teaser-content]()'>
<fig</div>
```

If I add two or more paragraphs, then everything works as expected.

Could you please check/fix it?

Thanks!

wordpress/wp-content/plugins/memberful-wp/src/global_marketing.php
```diff
@@ -0,0 +1,55 @@
...
+	remove_action( 'the_content', 'memberful_wp_protect_content', -10 );
+
+	$original_content = apply_filters( 'the_content', $post->post_content );
+
+	// re-add the action for follow-on call
+	add_action( 'the_content', 'memberful_wp_protect_content', -10 );
+
+	$offset = 0;
+	for ( $i = 0; $i < MEMBERFUL_PARAGRAPH_COUNT; $i++ ) {
+		$offset = strpos( $original_content, '</p>', $offset ) + 5;
```
northerner, line 37: Does this assume all content for posts will be in <p> tags?

I'm wondering how this would handle a post that is made up of content in a variety of other tags like `div`, `span`, `blockquote`, `video` etc. Or a post that has a lot of other tags before a couple of <p> tags at the end. Would that mean the offset includes all that other content in the teaser?

ChaseWPDEV: @northerner Correct. That is a deficiency in the way this feature currently functions, it really only works for long-form content. We could open up the project scope and requirements to allow the user to specify their tags -- but that would add some complexity and cost.

Alternately, we could fall back to using wp_excerpt, if that's preferable.

ChaseWPDEV:
@mrhead  Changes per feedback received made in e901a1d.

Based on the parameters that could result in a publisher not reaching the two-paragraph threshold I implemented this logic:
- If a post is protected, but does not reach the "2 paragraph" minimum for teaser content, then it only displays the global marketing content. If it does reach the minimum, then it displays the first two paragraphs, followed by the marketing content.

We can, of course, adjust that if you like. But I figured it was a good opening bid.

Lauren-Memberful:
Hey @ChaseWPDEV! Just letting you know that Patrik is OOO for a few days because of Easter.

wordpress/wp-content/plugins/memberful-wp/src/global_marketing.php
mrhead: I think that using `<p>` tags is OK for now. We can add support for other tags based on user feedback.

mrhead:
> Based on the parameters that could result in a publisher not reaching the two-paragraph threshold I implemented this logic:
> 
> - If a post is protected, but does not reach the "2 paragraph" minimum for teaser content, then it only displays the global marketing content. If it does reach the minimum, then it displays the first two paragraphs, followed by the marketing content.
> 
> We can, of course, adjust that if you like. But I figured it was a good opening bid.

I think that this could work for v1. Thanks!

mrhead:
> Changes per feedback received made in https://github.com/memberful/memberful-wp/commit/e901a1dcfb6717ac245a3909c8c42189df4ceae9.

@ChaseWPDEV It looks that the "max sessions" feature still doesn't work as expected. When I set the limit to two sessions, and try to log in three times from three different browsers, then sometimes only the latest session is logged out when I log in the third time, but sometimes both previous sessions are logged out, and I end up with only one session. Could you please look at it?

ChaseWPDEV:
@mrhead Did some interesting testing with Sessions and sign-ons. There was a condition where one Memberful Signon was resulting in 2x WordPress Signons for a single device.

The solution I've implemented is to only call the Memberful Authenticator signon, if called from a *.memberful.com referer. In my testing, this eliminated the problem of multiple sessions for one device and made the limits behave consistently.

mrhead:
Thanks, @ChaseWPDEV!

We've decided to put the session limiting feature on hold and will deploy the two other features in #320. We'll probably keep this PR open for now, but you can consider your work on this project done.

ChaseWPDEV:
@mrhead  Sounds good. And please do feel free to reach out if any questions come up on the session limiting. I'll be happy to see this through to the end -- even if it is at a much later date.

mrhead:
Thanks, @ChaseWPDEV! We appreciate your willingness to help.

When playing with session limiting, we noticed that it is super easy to log in back when you get logged out. From the technical point of view, everything works as expected. However, this doesn't prevent users from sharing credentials. It makes it just a little bit inconvenient for them. We would like to see the feature live, so we might revisit it later, but we are starting a new product cycle, so we are putting it on hold for now.

ChaseWPDEV:
Makes sense. FWIW, I'd say (market) that the feature "discourages" credential sharing (rather than preventing it). You can still share your creds, but you're going to get logged out every time your friend logs in, and vice-versa. I have to play this game occasionally with multiple admins trying to edit a WP post on a site and can vouch: the frustration is real.

okdan:
Just to clarify @ChaseWPDEV - this is unique to Memberful, because although the user is logged out of WP they remain logged into Memberful. So logging back in is much easier than a typical "log in" - it's essentially one-click. It's something we overlooked in the planning process, and might reevaluate later.

Lauren-Memberful:
Hey @mrhead! Tagging you in to review this one.

PR #319, opened 2022-04-14
Fix missing variable in compact call for ensure_user

julianfssen:
To-do thread:

[Possible Bug — errors from our WP plugin logged by the customer — APKMirror](https://3.basecamp.com/3293071/buckets/5610905/todos/4838922987)

This PR fixes a PHP deprecation warning when calling `compact` in the `ensure_present` function.

The related deprecations:

[PHP 7.3: Warning on undefined variable names passed to compact](https://php.watch/versions/7.3/compact-warn-undefined-vars)
[PHP 8.1: Warning on compact function calls with non-string and non-array string parameters](https://php.watch/versions/8.1/compact-non-string-warning)

However, my approach to fix this may be inaccurate as I'm not too familiar with PHP's conventions. I'll fix it if there's a better to address this.

### DIFF
--- a/wordpress/wp-content/plugins/memberful-wp/src/user/map.php
+++ b/wordpress/wp-content/plugins/memberful-wp/src/user/map.php
@@ -146,7 +146,11 @@ class Memberful_User_Mapping_Ensure_User {
     $user_id = $this->user_exists ? $this->update_user() : $this->create_user();
 
     if ( is_wp_error( $user_id ) ) {
-      $user_id->add_data( array_merge( (array) $user_id->get_error_data(), compact('user_data') ) );
+      if (isset($user_data)) {
+        $user_id->add_data( array_merge( (array) $user_id->get_error_data(), compact('user_data') ) );
+      } else {
+        $user_id->add_data( array_merge( (array) $user_id->get_error_data(), array() ) );
+      }
 
       return $user_id;
     }
--- a/wordpress/wp-content/plugins/memberful-wp/src/user/map.php
+++ b/wordpress/wp-content/plugins/memberful-wp/src/user/map.php
@@ -146,11 +146,7 @@ class Memberful_User_Mapping_Ensure_User {
     $user_id = $this->user_exists ? $this->update_user() : $this->create_user();
 
     if ( is_wp_error( $user_id ) ) {
-      if (isset($user_data)) {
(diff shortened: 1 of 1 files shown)


wordpress/wp-content/plugins/memberful-wp/src/user/map.php
```diff
@@ -146,7 +146,11 @@ public function ensure_present() {
     $user_id = $this->user_exists ? $this->update_user() : $this->create_user();
 
     if ( is_wp_error( $user_id ) ) {
-      $user_id->add_data( array_merge( (array) $user_id->get_error_data(), compact('user_data') ) );
+      if (isset($user_data)) {
```
mrhead, line 149: I wonder what scope `isset()` uses. `$user_data` is not set in this function's scope so I presume that `isset($user_data)` will never return `true`. I'll probably play with it later.

mrhead: `$user_data` used to be set in the function's scope, but https://github.com/memberful/memberful-wp/commit/bad84b0fd0331972a16f16c4bb3ab1ba14dccaba#diff-b9bd8651174701b4c1f2fdd53734864ae83e9cafd63235fa81124b1a49f60ffbL153 removed it.

Therefore, I presume we can get rid of the `compact()` call.

wordpress/wp-content/plugins/memberful-wp/src/user/map.php
```diff
@@ -146,7 +146,11 @@ public function ensure_present() {
...
+        $user_id->add_data( array_merge( (array) $user_id->get_error_data(), compact('user_data') ) );
+      } else {
+        $user_id->add_data( array_merge( (array) $user_id->get_error_data(), array() ) );
```
mrhead, line 152: I presume that this should work, too:

```suggestion
        $user_id->add_data( (array) $user_id->get_error_data() );
```

julianfssen: Thanks for pointing out the `user_data` variable! I was curious why I couldn't find it when searching, and I (incorrectly) assumed it was a WP or plugin method somewhere.

I've also updated the code to remove the `compact` and `array_merge`. It really didn't make sense to merge an empty array in the first place, not sure what I was thinking there 😅.

mrhead (approved):
Thanks! 👍

northerner (approved):

PR #328, opened 2022-09-26
Disconnect from Memberful app on uninstall

lmansur:
When disconnecting the integration (in WP) or uninstalling the plugin, call an endpoint in the Memberful app to destroy the integration. This ensures we stop sending webhooks to sites that no longer use the plugin.

Also resets the plugin's connection credentials.

https://3.basecamp.com/3293071/buckets/5610677/todolists/5361289350


mrhead (approved):
Nice!

PR #329, opened 2022-10-17
Use `docker compose` instead of `docker-compose`

mrhead:
`docker-compose` (AKA Docker Compose V1) is now deprecated and replaced by `docker compose` (AKA Docker Compose V2).

More info:
- https://docs.docker.com/compose/reference/
- https://www.docker.com/blog/announcing-compose-v2-general-availability/

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,8 +5,8 @@
 ### Setup instructions
 
 - Install [Docker](https://www.docker.com/get-started).
-- Run `docker-compose up` to start all needed containers. You can stop them with Ctrl+C.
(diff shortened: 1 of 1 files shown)


mrhead:
@lmansur `docker compose` is now available on Linux, too. Could you please verify it works for you?

https://docs.docker.com/compose/install/linux/

lmansur:
> @lmansur `docker compose` is now available on Linux, too. Could you please verify it works for you?
> 
> https://docs.docker.com/compose/install/linux/

Yeah, works great!

lmansur (approved):

julianfssen (approved):

PR #332, opened 2022-11-04
Update to Global Marketting feature

ChaseWPDEV:
- Updating  global marketing content settings
- Snippets are now optional
- Marketing content entry required (cannot be left blank when turned on)
- Migrates away from default content settings
- Per dropbox spec: https://www.dropbox.com/scl/fi/eaaknj0pootaiht4wq6l5/Memberful-WordPress-plugin-Updates-to-global-marketing-content.paper?dl=0&rlkey=n009qg03zyara3qgsvq5tly2l

### DIFF
--- a/wordpress/wp-content/plugins/memberful-wp/src/marketing_content.php
+++ b/wordpress/wp-content/plugins/memberful-wp/src/marketing_content.php
@@ -4,20 +4,20 @@ define( 'MEMBERFUL_MARKETING_META_KEY', 'memberful_marketing_content' );
 define( 'MEMBERFUL_LEGACY_DEFAULT_MARKETING_CONTENT', 'memberful_default_marketing_content' );
 
 function memberful_migrate_from_legacy_default(){
-  $legacy=get_option(MEMBERFUL_LEGACY_DEFAULT_MARKETING_CONTENT);
+  $legacy = get_option(MEMBERFUL_LEGACY_DEFAULT_MARKETING_CONTENT);
 
   if( empty($legacy) ){
     return;
   }
+
   //migrate to new settings
-  update_option('memberful_global_marketing_content', $legacy);
-  update_option('memberful_use_global_marketing', TRUE);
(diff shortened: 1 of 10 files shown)


mrhead:
@ChaseWPDEV Could you please update the PR title and description to make them more descriptive? They currently don't say anything about the changes in this PR.

wordpress/wp-content/plugins/memberful-wp/src/marketing_content.php
```diff
@@ -1,7 +1,23 @@
 <?php
 
 define( 'MEMBERFUL_MARKETING_META_KEY', 'memberful_marketing_content' );
-define( 'MEMBERFUL_OPTION_DEFAULT_MARKETING_CONTENT', 'memberful_default_marketing_content' );
+define( 'MEMBERFUL_LEGACY_DEFAULT_MARKETING_CONTENT', 'memberful_default_marketing_content' );
+
+
+
+function memberful_migrate_from_legacy_default(){
```
mrhead, line 8: Shouldn't we call this function from `memberful_wp_plugin_migrate_db()?` Currently, we don't call this function from anywhere, which means that the migration in the specification is not working.

mrhead (commented):
@ChaseWPDEV, thanks for the pull request.

I left one comment here and one in the specification box. Please check them both. Thanks!

wordpress/wp-content/plugins/memberful-wp/src/marketing_content.php
ChaseWPDEV: @mrhead  Good catch. I've fixed this with one modification -- I've added a hook to directly hook this function in to `admin_init` inside wordpress/wp-content/plugins/memberful-wp/src/marketing_content.php. 

Since there's no autoloader this saves us headaches from dependencies and `require` ordering.

mrhead: Thanks!

PR #333, opened 2022-11-14
Simplify DB reset instructions

mrhead:
I'm not sure why this was added to README.

`docker compose down` clears the database, so we do not need special instructions for it.


lmansur (approved):

julianfssen (approved):
Thanks, Patrik!

PR #338, opened 2022-12-13
Escape content before `echo`ing them

lmansur:
* afee47f **Set content-type for webhook response**

* 483b6dd **Escape content before `echo`ing them**
  
  The methods used in this commit are:
  * [`esc_attr`][1]: for content that will be used as an HTML attribute
  * [`esc_html`][2]: for content that will be placed inside an HTML tag
  * [`esc_url`][3]: for content that will be used inside URL-related HTML
    attributes, like `src` and `href`
  * [`wp_kses_post`][4]: for content that will be used inside a post (or
    in post-like context)
  * [`intval`][5]: converts values to `int`
  
  https://3.basecamp.com/3293071/buckets/9856127/todos/5621905896
  
  [1]: https://developer.wordpress.org/reference/functions/esc_attr/
  [2]: https://developer.wordpress.org/reference/functions/esc_html/
  [3]: https://developer.wordpress.org/reference/functions/esc_url/
  [4]: https://developer.wordpress.org/reference/functions/wp_kses_post/
  [5]: https://www.php.net/manual/en/function.intval

* cafa6f9 **Revert `intval($member_id)`**
  
  I didn't notice we were already casting to `int` with `(int)`

### DIFF
--- a/wordpress/wp-content/plugins/memberful-wp/src/cron.php
+++ b/wordpress/wp-content/plugins/memberful-wp/src/cron.php
@@ -18,7 +18,7 @@ function memberful_wp_cron_sync_users() {
 
   $members_to_sync = Memberful_User_Mapping_Repository::fetch_ids_of_members_that_need_syncing();
 
-  echo "<pre>library=memberful_wp fn=memberful_wp_cron_sync_users at=start members=".count($members_to_sync)."\n</pre>";
+  echo "<pre>library=memberful_wp fn=memberful_wp_cron_sync_users at=start members=".intval(count($members_to_sync))."\n</pre>";
 
   foreach ( $members_to_sync as $member_id ) {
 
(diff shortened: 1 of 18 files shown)


mrhead (approved):
Thanks @lmansur!

Could you please rebase this agains `data-sanitization-and-validation`? I did not realize that you based your branch on mine and I rebased mine against `main`, so this PR shows also changes made in my PR.

mrhead:
Or, if you don't mind, I can rebase it and merge it into my branch, too.

lmansur:
@mrhead done!

lmansur:
@mrhead I think we could keep this open until you merge the base PR tomorrow so the others have a chance to review it separately but I don't feel too strongly.

jcbpl (approved):

julianfssen (approved):
Thanks, Lucas! That's a _lot_ of escapes in Wordpress-land 😳

mrhead:
@lmansur I'm going to merge this in, so I can request the plugin review.

PR #339, opened 2022-12-14
Sanitize, escape, and validate data

mrhead:
https://3.basecamp.com/3293071/buckets/9856127/todos/5621901887

---

* ccca242 **Validate $_POST['memberful_product_acl'] and $_POST['memberful_subscription_acl'] in src/metabox.php**

* 396f9e6 **Sanitize $_POST['memberful_marketing_content'] in src/metabox.php**

* c676e62 **Sanitize and validate POST inputs in `memberful_wp_bulk_protect()`**

* 6a4b21b **Sanitize and validate POST inputs in `memberful_wp_protect_bbpress()`**

* 65fe53f **Validate POST inputs in `memberful_wp_private_rss_feed_settings()`**

* 1d71c95 **Sanitize values from `$_GET['category']`**

* c5d6b58 **Sanitize redirect URLs with `wp_sanitize_redirect()`**

* d0a52d7 **Sanitize `$_GET['member-feed']`**
  
  This seems unnecessary because we pass it to `WP_Query` which should
  sanitize it.
  
  However, I'm afraid that we need to do this anyway to make sure we pass
  the security audit.

### DIFF
--- a/wordpress/wp-content/plugins/memberful-wp/src/admin.php
+++ b/wordpress/wp-content/plugins/memberful-wp/src/admin.php
@@ -561,7 +561,10 @@ function memberful_wp_protect_bbpress() {
 
 function memberful_wp_private_rss_feed_settings() {
   if(isset($_POST['memberful_private_feed_subscriptions_submit'])) {
-    $private_feed_subscriptions = isset($_POST['memberful_private_feed_subscriptions']) ? $_POST['memberful_private_feed_subscriptions'] : false;
+    $private_feed_subscriptions = empty( $_POST['memberful_private_feed_subscriptions'] ) ? array() : (array) $_POST['memberful_private_feed_subscriptions'];
+    $private_feed_subscriptions = array_map( 'intval', $private_feed_subscriptions );
+    $private_feed_subscriptions = array_intersect( $private_feed_subscriptions, array_keys( memberful_subscription_plans() ) );
+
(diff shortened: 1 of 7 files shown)


wordpress/wp-content/plugins/memberful-wp/src/admin.php
```diff
@@ -414,48 +414,72 @@ function memberful_wp_advanced_settings() {
...
-      break;
-    case 'all_posts':
-      $query_params['post_type'] = 'post';
-      break;
-    case 'all_posts_from_category':
-      $query_params['category__in'] = $categories_to_protect;
-      break;
-    default:
-      $query_params['post_type'] = $things_to_protect;
+    if ( isset( $_POST['target_for_restriction'] ) ) {
```
lmansur, line 437: Non-blocking. WDYT of moving this check to the beginning of the method and calling `wp_die` if it's not set? I'm finding these nested conditions a bit hard to follow.

lmansur (approved):
Thank you for leading this ~annoying~ fun project! Got a tiny, non-blocking suggestion, but looks good!

wordpress/wp-content/plugins/memberful-wp/src/admin.php
mrhead: I slightly prefer to keep it as it is, so all `$_POST['target_for_restriction`]` checks are together.

But I'm open to refactoring it later.

jcbpl (approved):
Thank you for handling this!

PR #341, opened 2022-12-19
Sanitize and escape data

lmansur:


mrhead (approved):
Thank you for handling this, @lmansur!

Neverending story!

PR #343, opened 2023-01-18
Use environment variable as SVN username

julianfssen:
To-do: https://3.basecamp.com/3293071/buckets/9856127/messages/5716152173

This PR updates the plugin release script to use the `$MEMBERFUL_SVN_USERNAME` environment variable as the username when pushing to `svn`.

This allows us to use our Wordpress committer accounts to commit changes, instead of using the hardcoded plugin owner account which is discouraged by Wordpress.


lmansur:
@julianfssen is it possible to set up the SVN user in a local config file, similar to git? I think that would be preferable over an env var.

julianfssen:
Thanks for the feedback, @lmansur ! I like the config approach you mentioned, and found a way to do it via `svn` groups, which I've shared in the updated PR description. Let me know what you think of it!

lmansur (approved):
🚀

PR #345, opened 2023-01-30
Allow `iframe` tag when using `wp_kses_post`

julianfssen:
To-do: https://3.basecamp.com/3293071/buckets/5610905/card_tables/cards/5724292048

We use [`wp_kses_post`](https://developer.wordpress.org/reference/functions/wp_kses_post/) to sanitize editor content including the content we use in the bulk restrict access tool editor (metabox).

This plugin allows the `iframe` tag in `wp_kses_post` (as it is stripped by default), allowing our customers to embed videos in the bulk restrict access tool.

This is not the best solution as it basically defeats the purpose of `wp_kses_post` but it is a quick patch. Ideally, we can check for only embeds from trusted sources (YouTube, Vimeo, etc.) but it'll require some investigation to find out how long that takes and whether we want to work on it.

That said, I'm open to rejecting and closing this PR as well as if we feel this is too much of a security risk to include. Thank you!


mrhead:
You are right that this change defeats the purpose of `wp_kses_post()`.

What I noticed with the regular post content (I.e., not our marketing content). WordPress just saves the video URL and converts it into an iframe when generating content. Ideally, we should do the same thing when generating marketing content. However, I wasn't able to implement it while playing with this. I'll give it another try.

mrhead:
This works:

```diff
diff --git a/wordpress/wp-content/plugins/memberful-wp/src/marketing_content.php b/wordpress/wp-content/plugins/memberful-wp/src/marketing_content.php
index a7bea41..6e329dd 100755
--- a/wordpress/wp-content/plugins/memberful-wp/src/marketing_content.php
+++ b/wordpress/wp-content/plugins/memberful-wp/src/marketing_content.php
@@ -14,6 +14,9 @@ function memberful_marketing_content( $post_id ) {
     $marketing_content = memberful_term_marketing_content( $term );
   }
 
+  $embed = new WP_Embed();
+  $marketing_content = $embed->autoembed( $marketing_content );
+
   return apply_filters( 'memberful_marketing_content', $marketing_content );
 }
```

The [`autoembed()`](https://developer.wordpress.org/reference/classes/wp_embed/autoembed/) function embedsURLs that are on their own line in the original content.

So instead of using iframes, the customer would need to copy the video URL into the marketing content.

@julianfssen, do you think this would work for the customer?

julianfssen:
> This works:
> 
> ```diff
> diff --git a/wordpress/wp-content/plugins/memberful-wp/src/marketing_content.php b/wordpress/wp-content/plugins/memberful-wp/src/marketing_content.php
> index a7bea41..6e329dd 100755
> --- a/wordpress/wp-content/plugins/memberful-wp/src/marketing_content.php
> +++ b/wordpress/wp-content/plugins/memberful-wp/src/marketing_content.php
> @@ -14,6 +14,9 @@ function memberful_marketing_content( $post_id ) {
>      $marketing_content = memberful_term_marketing_content( $term );
>    }
>  
> +  $embed = new WP_Embed();
> +  $marketing_content = $embed->autoembed( $marketing_content );
> +
>    return apply_filters( 'memberful_marketing_content', $marketing_content );
>  }
> ```
> 
> The [`autoembed()`](https://developer.wordpress.org/reference/classes/wp_embed/autoembed/) function embedsURLs that are on their own line in the original content.
> 
> So instead of using iframes, the customer would need to copy the video URL into the marketing content.
> 
> @julianfssen, do you think this would work for the customer?

Thanks, @mrhead  The `autoembed` looks really clean and I think this should be fine for the customer, since it converts the link into an `iframe` similar to the one in share buttons for video platforms anyway. It also works out of the box for Vimeo _and_ YouTube so that's great.

I'll double check with Eyal and the customer on this but I think we can merge this in regardless since it seems like a good QOL improvement.

Thanks, Patrik!

mrhead (approved):
Thanks, @julianfssen!

PR #347, opened 2023-03-02
Allow iframes in marketing content

jcbpl:
This adds `iframe` to the HTML tags allowed by `wp_kses`, so that customers don't need to use a separate plugin to enable iframes in marketing content (e.g. for Spotify trailers).

It also matches the sanitization applied in the global content editor to the post and bulk restrict editors.


julianfssen (approved):
Thanks, @jcbpl ! I agree that since Gutenberg already allows `iframes`, we can allow it as well. I also like the idea of using a custom `wp_kses` function rather than overriding the original `wp_kses_post` tags which I tried previously. This is safer and also easier to understand.

I'm onboard with this change! Thanks, Jake!

mrhead (approved):

PR #354, opened 2024-04-03
Fix deprecated string interpolation syntax

mrhead:
The old syntax is deprecated since PHP 8.2.

See:
- https://php.watch/versions/8.2/%24%7Bvar%7D-string-interpolation-deprecated
- https://www.php.net/manual/en/language.types.string.php#language.types.string.parsing

Fixes: https://3.basecamp.com/3293071/buckets/5610905/card_tables/cards/7252740605


julianfssen (approved):
Thanks, @mrhead ! I'll have to keep this in mind next time 🤔

PR #358, opened 2024-04-16
Reduce number of tags in the SVN repository

mrhead:
This will speed up plugin releases because we checkout the whole repository every time we run the release script.

Fixes: https://3.basecamp.com/3293071/buckets/9856127/todos/7278826549

See also: https://developer.wordpress.org/plugins/wordpress-org/plugin-developer-faq/#how-many-old-releases-should-i-keep-in-svn

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -81,12 +81,11 @@ push_to_wordpress_svn() {
   svn copy trunk tags/$VERSION
   cd tags/$VERSION
   svn commit -m "$SVN_COMMIT_MESSAGE"
-}
 
-  # Keep only the latest 20 tags
+  # Keep only the latest 10 tags
   echo "Checking and removing old tags if necessary"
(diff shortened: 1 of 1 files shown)


jcbpl (approved):

release.sh
```diff
@@ -83,6 +83,18 @@ push_to_wordpress_svn() {
   svn commit -m "$SVN_COMMIT_MESSAGE"
 }
 
+  # Keep only the latest 20 tags
```
jcbpl, line 86: I think we could even reduce this to 2 or 3.

julianfssen (approved):
Thanks, @mrhead !

release.sh
mrhead: I wanted to be conservative :). Let's go with ten tags for now. We can lower it later if we need to.

PR #364, opened 2024-09-20
Pin WP docker image version to 6.4.0

lmansur:
WP updated Debian to 12 on versions >= 6.4.1, and it seems like it broke something on our container's DNS resolution—it can no longer reach the Rails app using puma-dev's subdomains (like apps.memberful.localhost)

I've spend a few hours trying to solve this but ultimately couldn't.

This commit pins the docker image version to the last one using Debian 11. Despite pinning WP to 6.4.0, `wp cli core install` installs the latest version, allowing us to test our plugin with the most recent WP.

We should find a long term fix for this, but for the time being, this will allow us to work on the plugin.

https://3.basecamp.com/3293071/buckets/9856127/todos/7308233397

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -17,7 +17,7 @@ services:
   wordpress:
     depends_on:
       - db
-    image: wordpress:latest
+    image: wordpress:6.4.0-php8.0
     ports:
       - "8181:80"
     restart: always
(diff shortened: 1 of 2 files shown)


mrhead (approved):
Thanks! I also spent a few hours trying to solve this, but to no avail.

PR #367, opened 2024-12-02
Fix `memberful_wp_user_disallowed_term_ids()`

mrhead:
The `memberful_wp_user_disallowed_term_ids()` function does not work as expected. It calls `memberful_wp_user_disallowed_ids_from_acl()`, which works well for posts but not for terms. This is because, instead of identifying all terms available to any registered users or subscribers, it relies on posts available to registered users and subscribers (see `$posts_for_any_registered_users` and
`$posts_for_anybody_subscribed_to_a_plan`).

Although `memberful_wp_user_disallowed_term_ids()` is buggy, it does not cause real-world issues. It is called from
`memberful_first_term_restricting_post()`, which checks for terms available to any registered users or subscribers. This workaround is necessary because `memberful_wp_user_disallowed_term_ids()` does not return these terms as expected.

This commit resolves the issue by removing the
`memberful_wp_user_disallowed_ids_from_acl()` function and integrating some of its logic directly into
`memberful_wp_user_disallowed_post_ids()` and
`memberful_wp_user_disallowed_term_ids()`. While this introduces some duplication, it improves the clarity and maintainability of both functions.

Additionally, I removed the static variables from both functions. These variables were intended to cache function calls, but no caching occurred due to a pre-declaration check on `$disallowed_post_ids` before marking it as `static`. I plan to revisit caching as I continue working on content restriction code, but for now, I prefer to exclude it for simplicity.

Also, I don't plan to release the on its own, so I'm not including a changelog entry.

### DIFF
--- a/wordpress/wp-content/plugins/memberful-wp/src/acl.php
+++ b/wordpress/wp-content/plugins/memberful-wp/src/acl.php
@@ -14,65 +14,68 @@ require_once MEMBERFUL_DIR . '/src/acl/term_options.php';
  * @return array Map of post ID => post ID
  */
 function memberful_wp_user_disallowed_post_ids( $user_id ) {
-  if ( isset( $disallowed_post_ids )) {
-    return $disallowed_post_ids;
-  } else {
-    static $disallowed_post_ids = array();
-    $acl = get_option( 'memberful_acl', array() );
-    $disallowed_post_ids = memberful_wp_user_disallowed_ids_from_acl( $user_id, $acl);
-
-    return $disallowed_post_ids;
-  }
-}
-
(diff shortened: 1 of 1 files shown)


mrhead:
Currently, `memberful_wp_user_disallowed_post_ids()` returns only posts disallowed by the post ACL, but not posts disallowed by assigned terms (e.g., categories). I plan to address this in a follow-up pull request so we can use `memberful_wp_user_disallowed_post_ids()` to exclude all disallowed posts from search results.

mrhead:
Closing in favor of https://github.com/memberful/memberful-wp/pull/368.

PR #368, opened 2024-12-03
Fix `memberful_wp_user_disallowed_term_ids()`

mrhead:
The `memberful_wp_user_disallowed_term_ids()` function does not work as expected. It calls `memberful_wp_user_disallowed_ids_from_acl()`, which works well for posts but not for terms. This is because, instead of identifying all terms available to any registered users or subscribers, it relies on posts available to registered users and subscribers (see `$posts_for_any_registered_users` and
`$posts_for_anybody_subscribed_to_a_plan`).

Although `memberful_wp_user_disallowed_term_ids()` is buggy, it does not cause real-world issues. It is called from
`memberful_first_term_restricting_post()`, which checks for terms available to any registered users or subscribers. This workaround is necessary because `memberful_wp_user_disallowed_term_ids()` does not return these terms as expected.

This commit fixes it by introducing _memberful_wp_items_from_acl(), which handles both posts and terms, ensuring consistent behavior. Additionally, function call caching is fixed; previously, `$disallowed_post_ids` was checked before being declared as `static`.

I don't plan to release the on its own, so I'm not including a changelog entry.

### DIFF
--- a/wordpress/wp-content/plugins/memberful-wp/src/acl.php
+++ b/wordpress/wp-content/plugins/memberful-wp/src/acl.php
@@ -84,15 +84,15 @@ function _memberful_wp_items_from_acl( $acl, $user_id, $items_for_subscribers, $
   $restricted_items = array_merge( $user_product_acl['restricted'], $user_subscription_acl['restricted'] );
 
   if ( $user_signed_in ) {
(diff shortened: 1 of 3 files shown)


lmansur (approved):
Nice work!

jcbpl (approved):
👍

PR #370, opened 2024-12-06
Improve teaser generation for global marketing content

mrhead:
Fixes:

- `strpos()` error for empty content
- Teaser generation for short content (e.g., just one paragraph)

See also: https://3.basecamp.com/3293071/buckets/5610905/card_tables/cards/8092300761

### DIFF
--- a/wordpress/wp-content/plugins/memberful-wp/src/global_marketing.php
+++ b/wordpress/wp-content/plugins/memberful-wp/src/global_marketing.php
@@ -52,25 +52,37 @@ function memberful_apply_global_snippets_content_filter( $memberful_marketing_co
   // re-add the action for follow-on call
   add_action( 'the_content', 'memberful_wp_protect_content', -10 );
 
-  $offset = 0;
-  for ( $i = 0; $i < MEMBERFUL_PARAGRAPH_COUNT; $i++ ) {
-    $offset = strpos( $original_content, '</p>', $offset ) + 5;
-    if( $offset === strlen($original_content) ){
-      continue;
-    }
-  }
-  $has_teaser= $offset < strlen($original_content);
+  $has_teaser = false;
+  $teaser = '';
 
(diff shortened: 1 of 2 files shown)


wordpress/wp-content/plugins/memberful-wp/src/global_marketing.php
```diff
@@ -52,25 +52,37 @@ function memberful_apply_global_snippets_content_filter( $memberful_marketing_co
...
+  if ( !empty( $original_content ) ) {
+    $teaser_offset = 0;
+
+    for ( $i = 0; $i < MEMBERFUL_PARAGRAPH_COUNT; $i++ ) {
+      $paragraph_offset = strpos( $original_content, '</p>', $teaser_offset );
+
+      if ( $paragraph_offset === false ) {
+        break;
+      } else {
+        $teaser_offset = $paragraph_offset + 4; // Move past the </p> tag
```
lmansur, line 67: Is the warning related to using `5` instead of `4`, which caused us to try and access an invalid position?

lmansur (approved):
Thank you!

wordpress/wp-content/plugins/memberful-wp/src/global_marketing.php
mrhead: No, I could only reproduce the warning with empty content. In the previous version of the code, we would perform two loop iterations with the empty content. During the first iteration, we added 5 to the offset, which caused `strpos()` to fail in the second iteration, as 5 exceeds the length of the content.

In fact, there were additional issues with the previous version of the code. I also don’t recall why we used 5 instead of 4 before. Furthermore, there are no visible changes in the original PR (https://github.com/memberful/memberful-wp/pull/318) from which I extracted this code. So, I guess we may never know.

jcbpl (approved):
🙌

PR #373, opened 2025-02-14
Disable the `bkismet_check_signup` filter on WordPress.com

mrhead:
This filter is automatically added to sites running on WordPress.com. It is designed to block signups that appear suspicious or spam-like; however, it also blocks legitimate email addresses.

Since the users we try to create in WordPress already exist in Memberful, it is safe to disable the filter when syncing users from Memberful to a WordPress site.

Fixes: https://3.basecamp.com/3293071/buckets/5610905/card_tables/cards/8294998242

### DIFF
--- a/wordpress/wp-content/plugins/memberful-wp/src/user/map.php
+++ b/wordpress/wp-content/plugins/memberful-wp/src/user/map.php
@@ -193,6 +193,8 @@ class Memberful_User_Mapping_Ensure_User {
   }
 
   private function create_user() {
+    global $wp_filter;
+
     $user_data = $this->fields_that_always_sync_from_memberful();
 
     $user_data['user_pass']               = wp_generate_password();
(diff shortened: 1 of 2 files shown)


wordpress/wp-content/plugins/memberful-wp/src/user/map.php
```diff
@@ -202,6 +204,22 @@ private function create_user() {
...
+          if ( is_array( $callback['function']) && is_object($callback['function'][0] ) ) {
+            $object = $callback['function'][0];
+
+            if ( method_exists( $object, 'bkismet_check_signup' ) ) {
+              remove_filter( 'wp_pre_insert_user_data', [$object, 'bkismet_check_signup'], $priority );
+            }
+          }
+        }
+      }
+    }
```
lmansur, line 221: Not blocking, but WDYT of extracting this to `core-ext.php` so we can share it with the rest of the plugin? I was thinking of something like:

`memberful_wp_remove_callback_from_filter('wp_pre_insert_user_data', 'bkismet_check_signup')`

Or will we run into trouble due to having to pass `$object` to some callbacks but not to others? I'm not familiar with what's inside `$callback`, so I don't know how simple this would be. I'll defer to you.

lmansur (approved):
Nice work tracking this down!

wordpress/wp-content/plugins/memberful-wp/src/user/map.php
mrhead: Good question. In general, it should be enough to call [remove_filter()](https://developer.wordpress.org/reference/functions/remove_filter/) directly because we know what `$callback` is.

`$callback` can be a function name, an array containing a class instance or a class name with a method, or an anonymous function. In this case, it is a class instance with a method name. However, since we don’t know the class instance, we need to iterate over all callbacks and remove the one where the method name matches.

I think this is a special case and not something we should be doing regularly, so I prefer keeping this code here rather than extracting it into a function.

lmansur: Makes sense, thanks!

jcbpl (approved):

PR #374, opened 2025-02-17
Refactor search protection logic in `memberful_wp_protect_search`

mrhead:
`pre_get_posts` can run (*) before WordPress is fully loaded, which may cause `current_user_can` to fail since it depends on `wp_get_current_user`, which might not be available yet. This issue occurs when plugins call `get_posts` before the init hook.

To prevent this, call `current_user_can` only after confirming that the query is a search query, ensuring that WordPress is fully loaded by then.

* For example, Oxygen Builder stores its templates as custom post types and it loads them with `get_posts` before WordPress fully initializes.

Fixes: https://3.basecamp.com/3293071/buckets/5610905/card_tables/cards/8325445010

### DIFF
--- a/wordpress/wp-content/plugins/memberful-wp/src/search_filter.php
+++ b/wordpress/wp-content/plugins/memberful-wp/src/search_filter.php
@@ -3,19 +3,30 @@
 add_action( 'pre_get_posts', 'memberful_wp_protect_search' );
 
 function memberful_wp_protect_search( $query ) {
-  // Do not filter search results for admins
-  if ( current_user_can( 'publish_posts' ) ) {
-    return $query;
+  // Do not protect admin pages.
+  if ( is_admin() ) {
+    return;
   }
 
-  if ( !is_admin() && $query->is_search() ) {
-    $disallowed_post_ids = memberful_wp_user_disallowed_post_ids( get_current_user_id() );
+  // Protect only search queries.
+  if ( !$query->is_search() ) {
+    return;
+  }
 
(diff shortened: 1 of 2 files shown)


jcbpl (approved):

wordpress/wp-content/plugins/memberful-wp/src/search_filter.php
```diff
@@ -3,19 +3,30 @@
 add_action( 'pre_get_posts', 'memberful_wp_protect_search' );
 
 function memberful_wp_protect_search( $query ) {
-  // Do not filter search results for admins
+  // Do not protect admin pages.
+  if ( is_admin() ) {
+    return;
+  }
```
lmansur, line 9: Similar to `current_user_can()`, can this fail because WP hasn't loaded yet? Should we do this after checking `is_search()`?

mrhead: It can’t. This function is defined in `wordpress/wp-includes/load.php` (one of the first required files), while `wp_get_current_user()` (which is called by `current_user_can()`) is defined in `wordpress/wp-includes/pluggable.php`, which is loaded later.

Also, the function name is misleading. `is_admin()` doesn’t check user capabilities; it checks whether the user is visiting a page in /wp-admin, so it doesn't need to access the current user.

lmansur: Great, thanks!

lmansur (approved):
Nice work!

PR #378, opened 2025-10-31
MR-3: User Roles Per Plan

zhickson:
Introduces per‑plan role mappings with settings UI and role-decision updates, plus ACL exclusivity/fallback tweaks and improved password reset role check.

Roles/Logic:

- Add per‑plan role mappings with storage (memberful_plan_role_mappings, memberful_use_per_plan_roles) and utilities (roles.php).
- Update Memberful_Wp_User_Role_Decision to assign roles from plan mappings, include fallbacks, capability-based priority, user cache clearing, and new filters.
- New helper memberful_wp_user_role_for_user() and use it in password‑reset audit.

Admin UI/Settings:

- Extend Advanced Settings to enable per‑plan roles and map plans → roles; optional bulk user role updates on save.
- New plan mappings view and URL helper; toggle UI visibility.

ACL/Access:

- Enforce mutual exclusivity between “any subscriber” and specific plan checks in post metabox (UI + save).
- Adjust post access logic to only apply “any subscriber” rule when no specific plan/product restrictions exist; otherwise honor intersections.

Hooks/Filters:

- Add filters for allowed roles, per‑plan usage, mappings retrieval, role decisions, and bulk update behavior.

### DIFF
--- a/wordpress/wp-content/plugins/memberful-wp/src/roles.php
+++ b/wordpress/wp-content/plugins/memberful-wp/src/roles.php
@@ -167,7 +167,7 @@ function memberful_wp_user_role_for_user( WP_User $user ) {
 
   $role_decision = Memberful_Wp_User_Role_Decision::build();
 
-  $user_role = $role_decision->role_for_user( memberful_wp_user_plans_subscribed_to( $user->ID ) );
+  $user_role = $role_decision->role_for_user( reset( $user->roles ), memberful_wp_user_plans_subscribed_to( $user->ID ) );
 
   /**
    * Filter to determine the user role for a user.
--- a/wordpress/wp-content/plugins/memberful-wp/src/user/role_decision.php
+++ b/wordpress/wp-content/plugins/memberful-wp/src/user/role_decision.php
@@ -51,14 +51,25 @@ class Memberful_Wp_User_Role_Decision {
 
   public function update_user_role( WP_User $user ) {
     $current_subscriptions = memberful_wp_user_plans_subscribed_to( $user->ID );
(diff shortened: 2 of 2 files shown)


wordpress/wp-content/plugins/memberful-wp/src/acl.php
```diff
@@ -295,11 +295,6 @@ function memberful_can_user_access_post( $user, $post ) {
     return $user ? true : false;
   }
 
-  // Grant access if user has a subscription and post or one of its terms allows access with any subscription
```
mrhead: Isn't it better to return early? I.e., if a post is viewable to any subscriber, isn't it better to check it at the beginning and return as soon as possible?

wordpress/wp-content/plugins/memberful-wp/src/roles.php
```diff
@@ -39,3 +49,136 @@ function memberful_wp_update_customer_roles( $old_active_role, $new_active_role,
...
+
+  foreach ( $users as $user ) {
+    Memberful_Wp_User_Role_Decision::ensure_user_role_is_correct( $user );
+  }
+}
+
+/**
+ * Get the assigned role for a user.
+ *
+ * TODO: Support for multiple roles.
```
mrhead: Could we remove this line? If we've decided not to support multiple roles yet but want to add it later, we should track that elsewhere, not in code comments.

mrhead (commented):
Nice work, @zhickson!

This looks good to me, and the feature works as expected. I've posted a few comments/questions here, but I also plan to post a few more in the related JIRA ticket.

PR #392, opened 2026-02-19
MR-9: Add membership expiry warning banner for members

zhickson:
## Summary
This PR introduces a new optional membership expiry banner to help members avoid unexpected access interruptions by surfacing renewal prompts before (and after) expiry.

When enabled, logged-in members see a front-end banner when at least one subscription is in scope (already expired, expiring today, or expiring within the configured reminder window). Messaging is context-aware across single and multiple subscriptions, including mixed states where some subscriptions are already expired while others expire soon.

## What's included
- New optional **Expiry Banner** feature in Memberful settings.
- Configurable reminder window (`1-90` days, default `7`).
- Front-end banner messaging for:
  - expired subscriptions
  - subscriptions expiring today
  - subscriptions expiring in N days
  - mixed subscription states (for example, some already expired and another expiring soon)
- Support for single- and multi-subscription accounts with accurate singular/plural wording.
- Auto-renew-aware filtering so eligible renewable subscriptions are not incorrectly treated as expiring.
- Dismiss control so members can hide the banner for the current session.
- Accessibility support (`role`/`aria-live`) and extension hooks for custom implementations (`memberful_expiry_banner_enabled`, `memberful_expiry_banner_days_threshold`, `memberful_expiry_banner_message`, `memberful_expiry_banner_aria_attributes`, `memberful_expiry_banner_html`).

## Scope and behaviour
- Shown only to logged-in users on front-end pages when criteria are met.
- Not shown in wp-admin.
- Not shown when the site is not connected or no qualifying subscriptions are found.
- No impact when the feature is disabled.


wordpress/wp-content/plugins/memberful-wp/src/options.php
```diff
@@ -24,6 +24,8 @@ function memberful_wp_all_options() {
     'memberful_filter_account_menu_items' => TRUE,
     'memberful_auto_sync_display_names' => FALSE,
     'memberful_show_protected_content_in_search' => FALSE,
+    'memberful_expiry_banner_enabled' => false,
```
mrhead:
```suggestion
    'memberful_expiry_banner_enabled' => FALSE,
```

We should probably use `FALSE` here to be consistent with the rest of the file.

wordpress/wp-content/plugins/memberful-wp/views/expiry-banner.php
```diff
@@ -0,0 +1,21 @@
...
+</style>
+<script async defer>
+  !function(){var e="memberful_expiry_banner_dismissed",n=document.getElementById("memberful-expiry-banner");if(!n)return;var t="memberful-expiry-banner-visible",o=document.getElementById("memberful-expiry-banner-bump-style");o||(o=document.createElement("style"),o.id="memberful-expiry-banner-bump-style",document.head.appendChild(o));var i=function(){var e="none"===n.style.display?0:n.offsetHeight;document.body&&document.body.classList.toggle(t,e>0),o.textContent="@media screen { html { margin-top: calc(var(--wp-admin--admin-bar--height, 0px) + "+e+"px) !important; } }"},r=function(){if(!window.sessionStorage)return!1;try{return"1"===window.sessionStorage.getItem(e)}catch(n){return!1}},s=function(){if(!window.sessionStorage)return;try{window.sessionStorage.setItem(e,"1")}catch(n){}};if(r())n.style.display="none";else{var a=n.querySelector(".memberful-expiry-banner__dismiss");a&&a.addEventListener("click",(function(){s(),n.style.display="none",i()}))}window.addEventListener("resize",i),window.addEventListener("orientationchange",i),i()}();
```
mrhead: @zhickson is there a reason this code is minified? It's pretty hard to read and understand. Also, is adding inline JS to plugins the best practise, or should we extract it?

wordpress/wp-content/plugins/memberful-wp/src/expiry_banner.php
```diff
@@ -0,0 +1,340 @@
...
+
+    if ( ! $is_expired && memberful_wp_subscription_has_autorenew_enabled( $subscription ) ) {
+      ++$active_subscriptions_count;
+      continue;
+    }
+
+    ++$expiring_subscriptions_count;
+
+    if ( null === $soonest || $expires_at < $soonest['expires_at'] ) {
+      $days_remaining = $is_expired ? 0 : (int) ceil( $seconds_remaining / DAY_IN_SECONDS );
```
mrhead: If I'm right, because we use `ceil`, this will always be 1 or greater for subscriptions that haven't expired yet. So we will never show the "Your subscription expires today" message, and instead we will show "Your subscription expires in 1 day" even if the subscription expires today.

Is that correct?

wordpress/wp-content/plugins/memberful-wp/views/expiry-banner.php
mrhead: Related: We should not use `async` and `defer` for inline scripts.

From: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/script#async

> Warning: This attribute must not be used if the src attribute is absent (i.e., for inline scripts) for classic scripts, in this case it would have no effect.

wordpress/wp-content/plugins/memberful-wp/src/expiry_banner.php
```diff
@@ -0,0 +1,340 @@
...
+
+/**
+ * Builds the user-facing banner message.
+ *
+ * @param array  $expiry_data Expiry information array.
+ * @param string $account_url Memberful account URL.
+ *
+ * @return string
+ */
+function memberful_wp_expiry_banner_message( array $expiry_data, $account_url ) {
```
mrhead, line 367: I've noticed two things in the banner message logic:

1. When all subscriptions are expiring (no active ones), `is_mixed_subscriptions` is false and the code falls through to "Your subscription has expired" (singular), even if multiple subs are expiring. Although this is an edge case which we can ignore if the fix is not trivial.

2. `memberful_wp_get_soonest_expiring_subscription` picks the smallest `expires_at`, so an already-expired sub always wins over one expiring tomorrow. The user sees "has expired" instead of the more actionable "expires in 1 day." Should not-yet-expired subs take priority?

mrhead (commented):
Thanks, @zhickson!

I've had a few comments. Could you please check them and let me know if my findings are correct?

Additionally, the PR description says "Not shown to admins." but https://github.com/memberful/memberful-wp/pull/392/changes/b37277325503d440df8e7fd0128454c4bef7476c removed this guard. Is that intentional?

wordpress/wp-content/plugins/memberful-wp/views/expiry-banner.php
zhickson: @mrhead we can extract the inline script, and have it minify as part of a build process. I used this approach to keep it as minimal code as possible, without introducing changes to the script build workflow, but I understand it does make it difficult to review.

Separately, I'll remove async/defer as you're right about it not being relevant to inline scripts.

wordpress/wp-content/plugins/memberful-wp/src/expiry_banner.php
zhickson: @mrhead correct. There are two main ways we could address this case:

1. Calendar-day: treat "today" as same local date (site timezone)
2. 24-hour window: treat anything < 24h as "today"

Do you have a preference here? I'd lean towards calendar day.

mrhead: Agreed, using calendar day should work here.

So if a subscription expires today, the banner will say "Your subscription expires today". If it expires tomorrow, the banner will say "Your subscription expires in 1 day".

wordpress/wp-content/plugins/memberful-wp/views/expiry-banner.php
mrhead: Thanks, @zhickson!

I'm unsure if we need to minify the script, as browsers and web servers usually use compression when transferring HTML, so the benefit of minifying it might be limited. Also, if the best practice is to use inline scripts for use cases like this one, we can keep the unminified inline script. But I'll leave this up to you.

mrhead (approved):
Nice work @zhickson and @av3nger!

PR #404, opened 2026-04-15
MR-7: Paywall Divider Block

zhickson:
**New Features**

- Added a Paywall Divider block to the editor that marks protected-content boundaries; content above the divider is shown as a preview to unauthorised visitors while content below remains protected.
- Editor now includes a Memberful inserter category for easy access to the block.

**Behaviour**

- Teaser and global marketing logic updated so posts with a divider render only the above-divider preview and adjusted settings text reflects this.

### DIFF
--- a/wordpress/wp-content/plugins/memberful-wp/src/content_filter.php
+++ b/wordpress/wp-content/plugins/memberful-wp/src/content_filter.php
@@ -2,6 +2,87 @@
 
 add_action( 'the_content', 'memberful_wp_protect_content', 100 );
 
+/**
+ * Split post content at the first paywall divider block.
+ *
+ * @param string $content Raw post content.
+ * @return array{
+ *   has_divider: bool,
+ *   content_above_divider: string,
+ *   content_below_divider: string
+ * }
+ */
+function memberful_wp_split_post_content_at_paywall_divider( $content ) {
+  $content = (string) $content;
+
+  if ( '' === $content ) {
+    return array(
+      'has_divider'            => false,
+      'content_above_divider'  => '',
+      'content_below_divider'  => '',
+    );
+  }
+
+  $divider_pattern = '/<!--\s+wp:memberful\/paywall-divider(?:\s+[^>]*)?\s*\/-->|<!--\s+wp:memberful\/paywall-divider(?:\s+[^>]*)?\s*-->\s*<!--\s+\/wp:memberful\/paywall-divider\s*-->/';
+  $content_parts   = preg_split( $divider_pattern, $content, 2 );
+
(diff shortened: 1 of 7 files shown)


wordpress/wp-content/plugins/memberful-wp/views/global_marketing.php
```diff
@@ -39,17 +39,18 @@
...
       <?php
       if ( $use_global_snippets ) :
         ?>
         checked="checked"<?php endif; ?>>
         <span class="memberful-label__text--multiline"><strong>Automatically pull an excerpt from each post.</strong>
-          <?php echo esc_html(' Memberful will pull the first two paragraphs from each protected post to use as marketing content for logged out visitors.'
-          .' This feature requires <p> tags in your posts to detect which content to use.');?>
+          <?php echo esc_html( ' Memberful will pull the first two paragraphs from each protected post to use as marketing content for logged out visitors.'
+          . ' This feature requires <p> tags in your posts to detect which content to use.'
+          . ' If a post contains a Memberful Paywall Divider block, all content above the divider will be shown instead of the excerpt.' ); ?>
```
mrhead, line 54: This is an pre-existing issue, but it looks like we should close the <span> tag here.

wordpress/wp-content/plugins/memberful-wp/src/content_filter.php
```diff
@@ -2,6 +2,87 @@
...
+    return $content;
+  }
+
+  if ( ! get_option( 'memberful_use_global_snippets' ) ) {
+    return $content;
+  }
+
+  $wrapped_content = "<div class='memberful-global-teaser-content'>$content</div>";
+
+  if ( function_exists( 'memberful_get_teaser_css' ) && ! did_action( 'memberful_teaser_css' ) ) {
```
mrhead:
```suggestion
  if ( function_exists( 'memberful_get_teaser_css' ) && ! did_filter( 'memberful_teaser_css' ) ) {
```

`memberful_teaser_css` is a filter, not an action, and calling `apply_filters` doesn't increment `did_action`.

mrhead (commented):
Thanks, @zhickson!

I left a comment regarding `did_action()`. I used an AI agent to help me with the review, but after checking the WP code, it looks like this issue is valid and we should probably use `did_filter()` instead.

zhickson:
@mrhead I've resolved the issues you mentioned above, and also slightly refactored how the paywall block is detected.
The reason for this is to prevent an additional call of `the_content` which I had previously. While in theory this shouldn't cause any issues, there is a small chance of regressions, and it's safer without it. So I refactored that slightly to avoid the extra call.

Any questions, please let me know.

mrhead:
Thanks, @zhickson! This looks good.

PR #416, opened 2026-06-19
Improve the debug tool

mrhead:
The debug tool renders all mapped and unmapped users. This is an issue on sites with many members because rendering the debug output fails.

This commit fixes it by removing the mapping info from the debug output, and adding an option to filter user/member mappings by either of email, `member_id`, or `wp_user_id`. In the UI this is a separate tab, so the debug tool is now split into system info and a member/user lookup. Over the debug endpoint the same `debug_tab` parameter selects the section, defaulting to a usage page that documents the available parameters.

Other small improvements:
- sensitive information is masked in system info and the error log
- error logs are HTML-escaped
- individual errors in the error log are separated so it's easier to scan

https://3.basecamp.com/3293071/buckets/9856127/todos/10000472282

### DIFF
--- a/wordpress/wp-content/plugins/memberful-wp/src/user/map_stats.php
+++ b/wordpress/wp-content/plugins/memberful-wp/src/user/map_stats.php
@@ -21,8 +21,7 @@ class Memberful_User_Map_Stats {
 
     return $wpdb->get_var(
       'SELECT COUNT(DISTINCT `mapping`.`wp_user_id`) FROM '.$this->table.' AS `mapping` '.
-      'INNER JOIN '.$wpdb->users.' AS `users` ON `users`.`ID` = `mapping`.`wp_user_id` '.
-      'WHERE `mapping`.`wp_user_id` > 0'
+      'INNER JOIN '.$wpdb->users.' AS `users` ON `users`.`ID` = `mapping`.`wp_user_id`'
     );
   }
 }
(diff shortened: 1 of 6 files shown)


wordpress/wp-content/plugins/memberful-wp/src/user/map_stats.php
```diff
@@ -8,30 +8,21 @@ public function __construct($table_name) {
...
+  // Joins against wp_users (and counts distinct ids) so orphaned and duplicate
+  // mappings don't inflate the figure beyond the number of real mapped users.
+  public function count_mapped_users() {
     global $wpdb;
 
-    return $wpdb->get_results('SELECT * FROM '.$this->table);
+    return $wpdb->get_var(
+      'SELECT COUNT(DISTINCT `mapping`.`wp_user_id`) FROM '.$this->table.' AS `mapping` '.
+      'INNER JOIN '.$wpdb->users.' AS `users` ON `users`.`ID` = `mapping`.`wp_user_id` '.
+      'WHERE `mapping`.`wp_user_id` > 0'
```
lmansur: I'm not following this part of the query:
```sql
'WHERE `mapping`.`wp_user_id` > 0'
```

Assuming it's to reject orphaned mappings, wouldn't the INNER JOIN do that already?

lmansur (approved):
This is great, thanks for doing it!

wordpress/wp-content/plugins/memberful-wp/src/user/map_stats.php
mrhead: Good catch. That's an artifact from an earlier version of the query. Removed in cba0b8ab213285a82bfac9accf2df50d30c6c653.

PR #417, opened 2026-06-22
MR-29: Copy update to gutenberg visibility settings

av3nger:
Update the wording of the visibility settings on the just-released gutenberg update to match what we have in the ad suppression visibility settings as well (which matches the metabox, etc). That way all of the language is consistent.


mrhead:
Thanks!

PR #419, opened 2026-06-22
MR-35: Paywall divider is misaligned

av3nger:
The paywall divider block should be aligned with the content of a post/page


mrhead:
Thanks!

PR #421, opened 2026-07-16
MR-36: Add WordPress Multisite support

okdan:
## Problem

Customers that runs two publications as subsites of one WordPress Multisite network, each connected to its **own Memberful account**. Multisite shares `wp_users`/`wp_usermeta` across the network, which broke the plugin in two coupled ways:

1. **Login bounce** — a member of both accounts signing in on the second site hit the anti-account-hijack guard (`user_already_exists`): the shared WP user existed, but the second site's mapping table had no entry, so the member was bounced to the WP password form for an account with no password they know.
2. **Entitlement clobbering** — entitlements were stored in plain usermeta keys (`memberful_subscription`, `memberful_product`, …) shared network-wide, so each site's sync overwrote the other's; content protection was wrong on both sites. This was masked only by bug 1 — fixing the bounce alone would have activated it, so both ship together.

## What changed

- **Blog-scoped member data**: new `memberful_wp_user_meta_key()` helper scopes per-account meta by blog ID on multisite (`memberful_product` → `memberful_2_product`, mirroring core's per-site `capabilities` pattern). All readers/writers routed through it: entity sync, ACL, expiry banner, private feed token (incl. the `WP_User_Query` lookup), custom field. **Single-site installs: keys and behavior unchanged.**
- **Migration (db_version 4)**: runs per-site on `admin_init`; copies existing plain-key values to the site's scoped keys for every user in that site's mapping table. Plain keys are left in place (other sites in the network may migrate later); any stale copies self-correct on the next routine member sync.
- **Network-aware login guard**: the `user_already_exists` bounce is skipped only when the colliding WP user is already Memberful-managed on another site in the network (checked via the other sites' mapping tables — OAuth has already proven email ownership). Genuine non-Memberful WP accounts still get the password verification.
- **Subsite membership**: sync now calls `add_user_to_blog()` when the member isn't yet a member of the current subsite, then refreshes the `WP_User` so the role decision sees the new role (without the refresh, `set_role(FALSE)` wiped the role).
- **Blog-scoped sync lock**: `memberful-member-mapping-{blog_id}-{member_id}` on multisite, so unrelated members that happen to share a numeric ID on different accounts can't contend or silently drop a webhook sync.
- **Per-site identity in the UI**: sync stores the member's name per site (`memberful_{blog_id}_full_name`); the profile widget shows the current site's name and only renders the signed-in block when the user is actually mapped on the current site (a shared network session no longer shows a false signed-in state with dead Account links).

The per-site `memberful_mapping` table was deliberately **kept** per-site: member IDs are only unique within one Memberful account, so a network-wide table can't represent multiple accounts.

## What's not included (deliberately)

Network-activation hook and a network admin settings page. Per-site activation and settings are the correct model here — each subsite connects its own account — and the mapping-table check gracefully skips sites where the plugin isn't active.

## Manual testing (Docker multisite, 3 sites)

- Two sites connected to two real local Memberful accounts; full OAuth sign-in on both sites with the same member email — no bounce, per-site mapping rows, fully isolated `memberful_1_*` / `memberful_2_*` meta, per-site ACL results.
- Guard regression: manually created WP user with a colliding email still triggers the password-verification flow.
- Migration: simulated legacy install (plain keys + db_version 3) migrates on `admin_init`; ACL reads work afterwards.
- Repeat syncs idempotent; third site without the plugin doesn't break the cross-site mapping check.
- Domain-mapped topology (matching production: one hostname per site) verified: independent sessions and sign-out per site.
- Single-site code path: scoping helper is a no-op (`is_multisite()` short-circuit), keys unchanged.

## Notes
- `readme.txt` changelog updated under `= unreleased =`; no version bump.
- Known limitation (accepted): shared WP profile fields (`first_name`/`last_name`) remain network-global, last-sync-wins. The plugin UI no longer depends on them; same email = same person, so divergence is rare.


zhickson (approved):
We've tested this on our end and happy with the first version of this feature

PR #431, opened 2026-08-17
MR-10: Allow setting paragraph count before paywall

av3nger:
# Summary

Customise how many paragraphs before paywall with global marketing content.

# What's included

- New memberful_paragraph_count option (1–10, default 2), registered in memberful_wp_all_options().
- New "Paragraphs before paywall" number field on the Global marketing content screen, shown only when "Automatically pull an excerpt" is enabled.
- The auto-excerpt teaser now reads the option (clamped 1–10) instead of the hardcoded MEMBERFUL_PARAGRAPH_COUNT constant, which is kept as the fallback default.

# Test plan

- On Settings → Memberful → Global marketing content, enable "Automatically pull an excerpt", set "Paragraphs before paywall" to N, save, and confirm the value persists (and that the field shows/hides with the excerpt checkbox).
- View a protected post with ≥N <p> paragraphs and no Paywall Divider block as a logged-out visitor, and confirm exactly N paragraphs appear before the paywall; posts using a Paywall Divider block are unaffected.
- Confirm out-of-range values clamp on save (0 → 1, 50 → 10).

### DIFF
--- a/wordpress/wp-content/plugins/memberful-wp/src/global_marketing.php
+++ b/wordpress/wp-content/plugins/memberful-wp/src/global_marketing.php
@@ -6,6 +6,7 @@ if ( ! defined( 'MEMBERFUL_PARAGRAPH_COUNT' ) ) {
 
 if(get_option('memberful_use_global_snippets')){
   add_filter( 'memberful_wp_protect_content', 'memberful_apply_global_snippets_content_filter', 1, 1 );
+  add_filter( 'memberful_wp_listing_excerpt', 'memberful_wp_apply_paragraph_count_to_listing_excerpt', 10, 2 );
 } else {
   add_filter( 'memberful_wp_protect_content', 'memberful_get_global_replacement', 1, 1 );
 }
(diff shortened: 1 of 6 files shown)


wordpress/wp-content/plugins/memberful-wp/src/global_marketing.php
```diff
@@ -75,8 +75,9 @@ function memberful_apply_global_snippets_content_filter( $memberful_marketing_co
 
   if ( !empty( $original_content ) ) {
     $teaser_offset = 0;
+    $paragraph_count = min( 10, max( 1, (int) get_option( 'memberful_paragraph_count', MEMBERFUL_PARAGRAPH_COUNT ) ) );
```
mrhead: This logic is repeated in three places.

Should we extract it somewhere?

```php
  function memberful_wp_paragraph_count(): int {
    return min( 10, max( 1, (int) get_option( 'memberful_paragraph_count', MEMBERFUL_PARAGRAPH_COUNT ) ) );
  }
```

mrhead (commented):
Thanks, @av3nger!

I've left one comment in Jira about the paragraph settings being ignored on
archive pages and one code-related comment here.

wordpress/wp-content/plugins/memberful-wp/src/global_marketing.php
av3nger: @mrhead, extracted to shared helper. The other issues have been addressed as well.

mrhead: Thanks!

PR #432, opened 2026-08-30
MR-11: Shortcodes to use BeaverBuilder shared rows just don't work

av3nger:
## Summary

This PR fixes a request-scoped side effect in the content protection filter that silently broke Beaver Builder layouts: when a protected post was rendered anywhere in a request (archive loop, widget, related-posts block), Memberful permanently removed Beaver Builder's the_content hook, so every BB layout rendered later in that same request disappeared and visitors saw raw fallback content instead.

While tracing that, several smaller defects in the same code path were found and fixed: a duplicate protection filter being registered while building the global marketing teaser, block-emitted shortcodes never executing inside marketing content, an empty teaser wrapper (plus its CSS) being output for posts with no "teasable" content, and the Sensei integration failing to unhook content protection at all.

##  What's included

- Beaver Builder hook restoration (src/content_filter.php): the gated branch still disables FLBuilder::render_content while the marketing/paywall content is built (so the protected layout can't leak into it), but now captures the hook's priority via has_filter() and re-registers it before returning. Posts rendered later in the same request get their layouts back.
- Duplicate filter removal (src/global_marketing.php): the teaser builder ran a stale remove_action / add_action pair at priority -10 around its nested the_content render. The removal was a silent no-op (the hook lives at priority 100), and the re-add registered a second protection filter at -10 for the rest of the request. Both lines are gone; recursion is already prevented by the existing doing_filter( 'memberful_wp_protect_content' ) guard.
- Block/shortcode ordering (src/content_filter.php): do_blocks on the marketing-content filter chain moved from priority 15 to 9, matching core's the_content ordering (blocks → wpautop → shortcodes). Shortcodes emitted by blocks in marketing content now execute instead of printing literally.
- Empty teaser suppression (src/global_marketing.php): $has_teaser was effectively always true, so protected posts with no extractable paragraphs emitted an empty memberful-global-teaser-content div plus the teaser gradient CSS. The teaser wrapper and CSS are now only output when there is actual teaser content.
- Sensei integration fix (src/contrib/woothemes-sensei.php): its remove_action( 'the_content', 'memberful_wp_protect_content' ) omitted the priority argument, making the removal a silent no-op since the hook is registered at priority 100. The priority is now passed.

##  Scope and behaviour

- The Beaver Builder change alters when the hook is absent, not what protected visitors see: the marketing/paywall content for the gated post itself renders exactly as before; only subsequent posts in the same request regain their layouts. The same capture/restore applies whatever priority BB registered at.
- No behaviour change for accessible posts, for admins (who bypass protection), or for sites without Beaver Builder - has_filter() returns false and both branches are skipped.
- The -10 cleanup and the $has_teaser fix only affect installs with global snippets enabled (memberful_use_global_snippets).
- The Sensei change makes the integration's intended existing behaviour (deferring protection to a custom course-aware filter) actually take effect; sites without Sensei are unaffected.

##  Test plan
  
All flows tested logged out on a site running Beaver Builder Pro, with global marketing + snippets enabled.

- Render a protected post followed by a public Beaver Builder page in the same request (e.g. an archive loop or a nested
  render). Before: BB's the_content hook is gone after the protected render and the BB page shows its raw fallback content. After: the hook survives (has_filter returns its original priority) and the BB layout renders.
- Protected BB page (regression): view a protected Beaver Builder page whose marketing content pulls in a BB shared row via [fl_builder_insert_layout] and a custom access-check shortcode. Logged out: teaser + marketing content + styled shared row, unchanged from before the fix. As a subscriber: full layout, no paywall markup.
- Empty teaser: view a protected post with no paragraph content. Before: page source contains an empty memberful-global-teaser-content div and the teaser <style> block. After: only the memberful-global-marketing-content div.
- Duplicate hook: after rendering a gated post with snippets enabled, dump the_content filter priorities. Before: memberful_wp_protect_content appears at both -10 and 100. After: only 100.
- Shortcodes in marketing content: marketing content containing a block that emits a shortcode now renders the shortcode output instead of the literal [shortcode] text.


wordpress/wp-content/plugins/memberful-wp/src/content_filter.php
```diff
@@ -217,8 +217,13 @@ function memberful_wp_protect_content( $content ) {
       return memberful_wp_listing_excerpt( $content, $content_split );
     }
 
-    // Disable Beaver Builder
-    remove_action( "the_content", "FLBuilder::render_content" );
+    // Disable Beaver Builder while the marketing content is built, so nested
+    // `the_content` calls can't render the protected layout into it.
+    $beaver_builder_priority = has_filter( 'the_content', 'FLBuilder::render_content' );
```
mrhead: Could we use the `fl_builder_do_render_content` filter instead of manually removing and then adding `FLBuilder::render_content` back?

The filter is available in the lite version of the plugin and based on the Pro 2.10.2.2
and Agency 2.9.4.2 leaks I found on GitHub it works there as well. Beaver Builder uses it internally too, for example in its WooCommerce Memberships compatibility code, so it looks like the intended way to keep the layout out of protected content.

By using the filter we don't need to remove the `FLBuilder::render_content`
hook and then add it back. Also the `$restore_beaver_builder` closure removes itself from `the_content` at priority 9999,
but that skips whatever is hooked at the next higher priority (for example Ad Inserter at 99999).

It could look like this:

```php
add_filter( 'fl_builder_do_render_content', function ( $do_render ) {
  if ( doing_filter( 'memberful_wp_protect_content' ) ) {
    return FALSE;
  }

  return $do_render;
} );
```

wordpress/wp-content/plugins/memberful-wp/src/contrib/beaver-builder.php
```diff
@@ -0,0 +1,290 @@
...
+}
+
+/**
+ * Apply the Memberful visibility rule when Beaver Builder decides whether to render a row, column, or module.
+ *
+ * @param bool   $is_visible Whether Beaver Builder considers the node visible.
+ * @param object $node       The node.
+ * @return bool Whether the node should be rendered.
+ */
+function memberful_wp_beaver_builder_is_node_visible( bool $is_visible, object $node ): bool {
```
mrhead: Non-blocking: Should we drop the parameter type declarations here? If another plugin's callback on this hook returns a wrong type (for example `null` because of a missing `return`), it fails here with a TypeError instead of in that plugin.

That said, this is definitely up to you since you have more experience than I do.

mrhead:
Thanks, @av3nger!

I had one suggestion and a small nit, otherwise this looks good to me. I'll
also post a question about the visibility rules to Jira.

wordpress/wp-content/plugins/memberful-wp/src/contrib/beaver-builder.php
av3nger: Dropped

wordpress/wp-content/plugins/memberful-wp/src/content_filter.php
av3nger: Done

mrhead:
Thanks, @av3nger!

I've merged this in a913f260 since there were merge conflicts in `readme.txt` that I needed to
fix.

PR #433, opened 2026-08-30
MR-15: Local dev env and CI/CD pipelines

av3nger:
# Summary

Modernises the plugin's tooling - repo layout, local dev environment, automated tests and linting. The plugin now lives at the repo root, local dev runs on `wp-env`, and CI covers tests, PHP compatibility, and coding standards.

# What's included

- **Layout + dev env:** plugin flattened to the repo root; Docker stack replaced with `@wordpress/env` (`.wp-env.json` + `env:*` npm scripts); `assets/` → `.wordpress-org/`.
- **Tests:** PHPUnit suite + `bin/install-wp-tests.sh`; `phpunit.yml` CI matrix (PHP 7.4–8.4 × WP, multisite, nightly).
- **Lint + compat:** `phpcs.xml` (WordPress standard, 2-space) + `.editorconfig`; `composer lint`/`compat` scripts; `lint.yml` with a **blocking** PHP 7.4+ compatibility gate and **advisory** coding-standards job (legacy backlog).
- **Pre-commit:** husky + lint-staged running `phpcs-changed` on staged PHP - checks only changed lines, so the legacy backlog never blocks a commit.

# Test plan

- `composer compat` → exits 0 (blocking gate, PHP 7.4+).
- `composer lint` → runs; legacy findings are advisory only (does not block).
- Pre-commit: staging a PHP file with a new error blocks the commit; merely touching a legacy file (clean change) commits fine. `--no-verify` bypasses.
- Open a PR → confirm **PHPUnit**, **PHP compatibility** (blocking), and **Coding standards** (advisory) jobs behave as expected.
- `npm run env:start` to init a local environment using Docker

On latest node versions, if the wp-env command fails with `ETIMEDOUT` error, run the following command before `npm run env:start`:
`export NODE_OPTIONS="--network-family-autoselection-attempt-timeout=5000 --dns-result-order=ipv4first"`

PR #434, opened 2026-08-30
MR-15: Release pipeline

av3nger:
**!! IMPORTANT !!** Depends on https://github.com/memberful/memberful-wp/pull/433

# Summary

Releases to WordPress.org are automated from GitHub Releases.

# What's included

- **Release pipeline:** `.distignore` (ships only the plugin; keeps `vendor/reporting.php`); `deploy.yml`; `build-zip.yml` (manual `workflow_dispatch` to build a test zip from any branch).

# Test plan

- Actions → **Build release zip** → run on any branch → unzip the artifact and confirm it contains `src/`, `views/`, `js/build/`, `vendor/reporting.php` and excludes `node_modules/`, `tests/`, `js/src/`, dev configs.
- Deploy (release only): add `SVN_USERNAME`/`SVN_PASSWORD` secrets; tag with the bare version (e.g. `1.78.0`) - a mismatched tag fails the version check. Optionally `dry-run: true` once before a real release.


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,38 @@
...
+    steps:
+      - uses: actions/checkout@v7
+      - name: Check release tag matches version
+        run: |
+          v=$(grep -oiP '^Stable tag:\s*\K\S+' readme.txt)
+          [ "$v" = "$GITHUB_REF_NAME" ] || { echo "::error::Release tag $GITHUB_REF_NAME != Stable tag $v"; exit 1; }
+
+      - name: Build
+        run: |
+          npm install
```
mrhead: Could we use `npm ci` here (and in `build-zip.yml`)? It installs exactly what `package-lock.json` says and fails if the lockfile is out of sync, so the release is built with the same dependency versions we tested with.

.github/workflows/build-zip.yml
```diff
@@ -0,0 +1,22 @@
...
+  build:
+    name: Build release zip
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+
+      - name: Build plugin
+        run: |
+          npm install
```
mrhead: Same here, `npm ci` instead of `npm install`.

package.json
```diff
@@ -6,7 +6,16 @@
   "scripts": {
```
mrhead, line 5: The `version` on line 3 (1.78.1) is out of sync with the plugin (1.81.0 in `readme.txt` and `memberful-wp.php`). Could we bump it, or drop the field if we don't use it?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,38 @@
...
+          npm run build
+
+      - name: WordPress plugin deploy
+        id: deploy
+        uses: 10up/action-wordpress-plugin-deploy@stable
```
mrhead: I'm not sure what the best practice is, but could we pin these actions (`10up/action-wordpress-plugin-deploy`, `softprops/action-gh-release` and `10up/action-wordpress-plugin-build-zip` in `build-zip.yml`) to commit SHAs to eliminate supply chain attacks? The deploy action will have access to the SVN credentials, so I would like to minimize the risk of leaking them.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,38 @@
...
+
+jobs:
+  deploy:
+    name: New release
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      - name: Check release tag matches version
+        run: |
+          v=$(grep -oiP '^Stable tag:\s*\K\S+' readme.txt)
```
mrhead: Could we extend this to also verify the `Version` header and the `MEMBERFUL_VERSION` constant in `memberful-wp.php`, not just the `Stable tag` in `readme.txt`? That's what `release.sh` did.

mrhead (commented):
Nice work, @av3nger!

I've left a couple of questions here and will also post some to JIRA.

av3nger:
@mrhead , feedback addressed

mrhead (approved):
Nice work!

mrhead:
Closing in favor of #443.

PR #435, opened 2026-08-31
MR-5: Metered Paywall

av3nger:
## Summary

This PR adds a metered paywall on top of the existing Memberful paywall: a "free article allowance" that lets anonymous visitors and registered free members read a configurable number of matching posts within a rolling period before the paywall is shown.
  
## What's included

- New Memberful - Metering settings tab: enable toggle, rolling period, anonymous limit, registered free member limit, and an "apply to protected posts" option.
- Visual rule builder. A post is metered when any rule group matches:
    - post type: is any of / is none of
    - category / tag: has any of / has none of
    - URL: contains / does not contain
- Server-side engine: decision cached per request; paid members, exempt posts, non-matching posts, and ineligible requests (admin, feed, REST, non-GET, authors) are never metered.
- Anonymous views tracked in a signed, HttpOnly cookie; registered views in user meta; anonymous views merge into the account on login.
- Per-post "Exempt from metering" metabox.
- Metering countdown block with a configurable `{count}` template, shown only while the visitor is still being sampled.

## Test plan

- Open Memberful - Metering, enable it, set a short period, an anonymous limit (e.g. 2) and a registered limit (e.g. 3), add a rule (e.g. post type is any of: Post), and save.
- As a logged-out visitor, view matching posts past the limit:
    - confirm the first N render in full and the next shows the paywall
    - confirm re-opening an already-read post still renders (no double count)
    - confirm a non-matching post is never metered
- Log in as a free member with prior anonymous views - confirm they carry over and the registered limit applies.
- As a paid member, confirm matching posts are never metered.
- Tick "Exempt this post from metering" and confirm the post stays fully readable and uncounted.
- Insert the countdown block, use the `{count}` button, and confirm it shows the remaining count while sampling and disappears once the meter trips.
- Disable metering and confirm no metering output or cache headers on protected posts.


wordpress/wp-content/plugins/memberful-wp/src/metering/access.php
```diff
@@ -0,0 +1,362 @@
...
+    if (
+      ! self::post_matches_any_group( $post, $config['rules'] )
+      || self::post_matches_any_group( $post, $exclude_rules )
+    ) {
+      self::cache( $post->ID, self::DECISION_IGNORE );
+      return;
+    }
+
+    $user_id  = get_current_user_id();
+    $has_paid = $user_id && (
```
mrhead: Could we rename `$has_paid` to `$has_plan_or_download`? That name is better because "paid" doesn't work
well with free trials and 100% coupons.

wordpress/wp-content/plugins/memberful-wp/src/content_filter.php
```diff
@@ -158,6 +158,103 @@ function memberful_wp_first_paragraphs( string $content, int $count ): string {
...
+  $blocks = memberful_wp_blocks_above_divider( parse_blocks( (string) $post->post_content ), $found );
+
+  // Fail closed: without the divider in the parsed tree there is nothing to cut at, so render no teaser rather than
+  // the whole post.
+  if ( ! $found || empty( $blocks ) ) {
+    return '';
+  }
+
+  // Truncated wrappers render without their closing markup (see memberful_wp_inner_content_for_blocks()).
+  return force_balance_tags( strip_shortcodes( do_blocks( serialize_blocks( $blocks ) ) ) );
```
mrhead: Would it be possible to use `render_block()` instead of `do_blocks()`?

When a site has classic posts (written without blocks), it breaks their paragraphs: `do_blocks()` unhooks `wpautop` from `the_content` and schedules its restore at priority 11, which never runs when we call it from priority 100, so the next post rendered with full content on the same page loses `wpautop`.

Example:

```php
$output = '';
foreach ( $blocks as $block ) {
  $output .= render_block( $block );
}

return force_balance_tags( strip_shortcodes( $output ) );
```

Would this work?

wordpress/wp-content/plugins/memberful-wp/src/metering/access.php
```diff
@@ -0,0 +1,362 @@
...
+    $already_counted = isset( $views[ $post->ID ] );
+
+    if ( ! $already_counted && count( $views ) >= $limit ) {
+      self::emit_no_cache_headers();
+      self::cache( $post->ID, self::DECISION_TRIP_METER, 0 );
+      return;
+    }
+
+    if ( ! $already_counted ) {
+      $views[ $post->ID ] = time();
```
mrhead, line 127: Would it be possible to disable metering for prefetch requests? Otherwise, in theory, it
could happen that the browser or a plugin would prefetch and trip the meter before a user
even reads anything.

Something like this should work:

```php
// In on_template_redirect(), right before a new view is recorded.
if ( ! $already_counted && self::is_prefetch_request() ) {
  // Browsers drop a failed prefetch and load the page normally on click, which is then counted.
  nocache_headers();
  status_header( 503 );
  exit;
}

private static function is_prefetch_request(): bool {
  foreach ( array( 'HTTP_SEC_PURPOSE', 'HTTP_PURPOSE', 'HTTP_X_MOZ', 'HTTP_X_SEC_PURPOSE' ) as $header ) {
    if ( isset( $_SERVER[ $header ] ) && preg_match( '/prefetch|prerender/i', (string) $_SERVER[ $header ] ) ) {
      return true;
    }
  }

  return false;
}
```

wordpress/wp-content/plugins/memberful-wp/src/metering/access.php
```diff
@@ -0,0 +1,362 @@
...
+        $post_terms = self::term_slugs_for_post( $post->ID, $taxonomy );
+        $matches    = ! empty( array_intersect( $values, $post_terms ) );
+        break;
+
+      case 'url':
+        $path    = wp_parse_url( (string) get_permalink( $post->ID ), PHP_URL_PATH );
+        $path    = is_string( $path ) ? $path : '';
+        $matches = false;
+        foreach ( $values as $fragment ) {
+          if ( '' !== $fragment && false !== strpos( $path, (string) $fragment ) ) {
```
mrhead: Currently we compare URLs case-sensitively. Should we use the case-insensitive `stripos()` instead?

For example, a rule "URL contains `/News/`" never matches, because WP slugs are always lowercase (`/news/`).

wordpress/wp-content/plugins/memberful-wp/src/paywall/renderer.php
```diff
@@ -205,14 +248,28 @@ private static function features_block( array $config ): string {
...
   private static function primary_cta( array $config, bool $interactive ): string {
-    $url = self::subscribe_url( $config );
+    $label = $config['button_label'];
+    $url   = self::subscribe_url( $config );
+
+    if ( self::should_show_free_registration_cta() ) {
+      $free_url = self::free_registration_url( $config );
+
+      if ( '' !== $free_url ) {
+        $label = $config['free_button_label'];
```
mrhead, line 277: Could we check that the label is not empty during render time for both `button_label` and `free_button_label`? Otherwise the paywall renders a button with no text.

wordpress/wp-content/plugins/memberful-wp/views/metering/metabox.php
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * @package memberful-wp
+ *
+ * @var bool $exempt Whether this post is currently exempt from metering.
+ */
+?>
+<label>
+  <input
+    type="checkbox"
+    id="memberful_metering_exempt"
```
mrhead: The metabox and the checkbox have the same ID (`memberful_metering_exempt`), which is technically invalid HTML. It looks like the checkbox doesn't need an ID, so we could remove it from there.

wordpress/wp-content/plugins/memberful-wp/src/metering.php
```diff
@@ -0,0 +1,52 @@
...
+Memberful_Metering_Metabox::register();
+
+/**
+ * Render and save the metering settings screen.
+ */
+function memberful_wp_metering_settings() {
+  if ( isset( $_POST['save_metering'] ) && memberful_wp_valid_nonce( 'memberful_options' ) ) {
+    $raw_config = empty( $_POST['memberful_metering'] ) ? array() : (array) wp_unslash( $_POST['memberful_metering'] );
+
+    Memberful_Metering_Config::save( $raw_config );
```
mrhead, line 24: Saving the metering settings doesn't provide any feedback. The same goes for the Global Paywall tab.

Could we fix that in both tabs? For example, the Hide ads tab shows the "Settings updated" notice after saving.

wordpress/wp-content/plugins/memberful-wp/js/src/blocks/metering-countdown/edit.js
```diff
@@ -0,0 +1,157 @@
...
+import { __ } from '@wordpress/i18n';
+
+/**
+ * React hook that is used to mark the block wrapper element.
+ * It provides all the necessary props like the class name.
+ *
+ * @see https://developer.wordpress.org/block-editor/reference-guides/packages/packages-block-editor/#useblockprops
+ */
+import {
+	BlockControls,
```
mrhead: In the .js files in main we use 2 spaces for indentation, but here and in `index.js` and `render.php` we use tabs.

Should we fix that for the sake of consistency?

mrhead (commented):
Nice work @av3nger! I had a few comments, but overall this looks good.

wordpress/wp-content/plugins/memberful-wp/src/metering/access.php
av3nger: Done

wordpress/wp-content/plugins/memberful-wp/src/content_filter.php
av3nger: Done

wordpress/wp-content/plugins/memberful-wp/src/metering/access.php
av3nger: Done

wordpress/wp-content/plugins/memberful-wp/src/metering/access.php
av3nger: Done

wordpress/wp-content/plugins/memberful-wp/src/metering.php
av3nger: Done

wordpress/wp-content/plugins/memberful-wp/views/metering/metabox.php
av3nger: Done

wordpress/wp-content/plugins/memberful-wp/js/src/blocks/metering-countdown/edit.js
av3nger: Done

av3nger:
@mrhead , I have addressed all the issues

mrhead:
Thanks, @av3nger! However, it looks like you forgot to push your changes as there are no new commits :).

av3nger:
@mrhead , sorry for that. I wasn't expecting you to see my message this fast :-)

mrhead (approved):
Thanks @av3nger!

I think this is looking good and is ready to merge.

I still want to double-check a few things about the prefetch handling I suggested yesterday, but I'll do that once I merge this PR.

mrhead:
@av3nger I've pushed a follow-up commit e7a288a1 removing the `X-Purpose` header from the prefetch check.

Safari no longer sends `X-Purpose`, and its known senders today are in-app browsers such as Facebook's, which may cache the 503 and show an error instead of the article. The other prefetch headers still get the 503 as before.

Additionally, the check now matches only a leading `prefetch` token, `emit_no_cache_headers()` sends the full no-cache directives on WordPress before 6.8, and the developer section was removed from `readme.txt`.

PR #436, opened 2026-09-01
MR-13: WP Recipe Maker compatibility

av3nger:
## Summary

This PR adds WP Recipe Maker compatibility to the paywall: a per-post "Lock recipe cards" option that keeps the recipe card visible as a teaser (name, image, summary, ratings, times, sharing buttons) while stripping all protected recipe details - ingredients, instructions, equipment, notes, nutrition, and video - from every output channel WPRM exposes to non-members.

##  What's included

- Per-post lock option. A "Recipe cards" checkbox in the Memberful metabox (shown only when WP Recipe Maker is active) locks the recipe cards in that post. Paid members with access, and users who can publish posts, always see the full card.
- Pruned recipe card teaser. Locked full recipe cards are pruned via DOM filtering rather than hidden entirely: teaser-safe components (identity, image, summary, meta, sharing, roundup/list layout) stay visible, protected sections are removed, and empty template wrappers are collapsed so the card doesn't leave behind blank layout containers.
- Component-level filtering. Individual WPRM blocks and wprm-* shortcodes (e.g. ingredients or instructions placed outside the card) are filtered through an allowlist, extensible via the memberful_wprm_teaser_visible_components filter.
- Leak-proofing across WPRM's side channels:
  - JSON-LD recipe metadata is emptied for locked recipes, and WPRM's metadata cache is disabled for protected recipes so the filter always runs.
  - The footer recipe JSON aggregate excludes locked recipes (resolved via the queried post, or the recipe's parent post outside singular views).
  - Print pages return nothing for locked recipes.
  - REST responses for the recipe post type have the recipe data and rendered content stripped (WPRM force-enables show_in_rest).
- Teaser integration. Locked recipe previews are appended to the teaser/marketing content on protected posts, unless the content already contains a recipe card.
- New core extension points (used by the integration, available to others): memberful_metabox_view_vars filter, memberful_save_postdata action, and a memberful_teaser_content filter applied through the new memberful_wp_wrap_teaser_content() helper.

##  Test plan

1. With WP Recipe Maker active, edit a post containing a recipe card, protect it with Memberful, and tick "Lock WP Recipe Maker recipe cards in this post."
2. As a logged-out visitor:
  - confirm the recipe card shows name, image, summary, and meta but no ingredients, instructions, notes, or nutrition
  - view the page source and confirm no recipe JSON-LD and no ingredient/instruction text in the footer recipe JSON
  - open the recipe's print URL and confirm it renders nothing
  - request the recipe via the REST API and confirm recipe data and rendered content are empty
3. As a member with access (and as an admin), confirm the full card, metadata, print page, and REST response are intact.
4. Confirm a post with the lock unticked behaves exactly as before, and that a locked recipe placed via individual blocks/shortcodes (outside the full card) is also hidden.
5. Confirm posts without recipes and sites without WP Recipe Maker are unaffected (integration only loads when the plugin is active).


wordpress/wp-content/plugins/memberful-wp/src/contrib/wp-recipe-maker.php
```diff
@@ -0,0 +1,893 @@
...
+
+  /**
+   * Register REST hooks once the WPRM post type constant is available.
+   */
+  public function register_rest_hooks(): void {
+    if ( ! defined( 'WPRM_POST_TYPE' ) ) {
+      return;
+    }
+
+    add_filter( 'rest_prepare_' . WPRM_POST_TYPE, array( $this, 'filter_rest_prepare_recipe' ), 100, 2 );
```
mrhead, line 365: The REST leak-proofing covers core's `wp/v2/wprm_recipe` route, but WP Recipe Maker also has its own public route, `POST /wp-json/wp-recipe-maker/v1/manage/recipe/{id}`, with no permission check. It returns the full ingredient list of a locked recipe to anyone. WPRM's front-end script calls it as a fallback whenever the recipe is missing from the footer JSON, which is exactly what happens for locked recipes.

Could we hook `wprm_recipe_frontend_data` and empty the ingredients when the recipe is locked for the current user? WPRM builds both this response and the footer JSON through that filter, so it would cover both.

```php
add_filter( 'wprm_recipe_frontend_data', array( $this, 'filter_recipe_frontend_data' ), 10, 2 );

public function filter_recipe_frontend_data( $data, $recipe ) {
  if ( is_array( $data ) && $this->should_lock_recipe_for_current_user( $recipe ) ) {
    $data['ingredients'] = array();
  }

  return $data;
}
```

wordpress/wp-content/plugins/memberful-wp/src/contrib/wp-recipe-maker.php
```diff
@@ -0,0 +1,893 @@
...
+    }
+
+    $data = $response->get_data();
+
+    if ( ! is_array( $data ) ) {
+      return $response;
+    }
+
+    if ( array_key_exists( 'recipe', $data ) ) {
+      $data['recipe'] = null;
```
mrhead, line 394: This strips the `recipe` field and the content, but the taxonomy fields stay in the response. WPRM assigns every ingredient and piece of equipment to the `wprm_ingredient` and `wprm_equipment` taxonomies, and both are exposed via REST. So a locked recipe still returns the term IDs, and `GET /wp-json/wp/v2/wprm_ingredient?post={id}` turns them into the ingredient names.

Could we empty `wprm_ingredient` and `wprm_equipment` here as well? Course, cuisine and keywords can stay, since the pruned card shows them anyway.

```php
foreach ( array( 'wprm_ingredient', 'wprm_equipment' ) as $taxonomy ) {
  if ( isset( $data[ $taxonomy ] ) ) {
    $data[ $taxonomy ] = array();
  }
}
```

wordpress/wp-content/plugins/memberful-wp/src/contrib/wp-recipe-maker.php
```diff
@@ -0,0 +1,893 @@
...
+    }
+
+    return WPRM_Recipe_Manager::get_recipe( $recipe_id );
+  }
+
+  /**
+   * Determine which post controls locking for a recipe.
+   *
+   * Requests without an embedding post context (print URLs, REST) resolve via
+   * the recipe's WPRM parent post; reuse in other posts is not scanned.
```
mrhead, line 671: Outside the post (print page, REST, the recipe's own URL) the lock resolves through the recipe's single parent post. That means a recipe created in a public post and reused in a locked post is fully accessible on those channels, while the card in the locked post looks locked. The code comment above says as much, but site owners won't see it.

Could we add a second description line under the checkbox, so people know the limitation? Something like:

> Recipes are protected through the post they were created in. A recipe reused in another post is not protected there.

wordpress/wp-content/plugins/memberful-wp/src/contrib/wp-recipe-maker.php
```diff
@@ -0,0 +1,893 @@
...
+   */
+  public function add_locked_recipe_preview_to_protected_content( string $content ): string {
+    static $adding_preview = false;
+    global $post;
+
+    if ( $adding_preview || ! $post instanceof WP_Post || memberful_can_user_access_post( get_current_user_id(), $post->ID ) ) {
+      return $content;
+    }
+
+    if ( get_option( 'memberful_use_global_snippets' ) && has_block( 'memberful/paywall-divider', $post ) ) {
```
mrhead: This renders the pruned card twice when global snippets are off and the card sits above the paywall divider. Steps: turn off global snippets, add a recipe card above the divider in a protected post, tick the lock, view logged out.

The card above the divider is already in the teaser, but this guard only skips the post when snippets are on, so the preview is inserted again before the marketing content. The divider path also skips the `memberful_teaser_content` filter when snippets are off, so it can't take over.

Suggested fix: let the divider path always run the teaser filter, and skip here whenever the post has a divider.

```php
// content_filter.php
function memberful_wp_format_divider_teaser_content( $content ) {
  if ( ! get_option( 'memberful_use_global_snippets' ) ) {
    return apply_filters( 'memberful_teaser_content', $content, $GLOBALS['post'] );
  }
  ...

// wp-recipe-maker.php
if ( has_block( 'memberful/paywall-divider', $post ) ) {
  return $content;
}
```

wordpress/wp-content/plugins/memberful-wp/src/contrib/wp-recipe-maker.php
```diff
@@ -0,0 +1,893 @@
...
+    }
+
+    $lock_post_id = $this->lock_post_id_for_recipe( $recipe, $post_id );
+
+    if ( ! $lock_post_id ) {
+      return false;
+    }
+
+    // A metered sample view releases the whole post body, recipe cards included.
+    if ( function_exists( 'memberful_metering_is_releasing' ) && memberful_metering_is_releasing( $lock_post_id ) ) {
```
mrhead: This looks like a leftover. `memberful_metering_is_releasing()` doesn't exist on main, so the guard is always false.

Could we merge main and then use `Memberful_Metering_Access` instead?

```php
if ( class_exists( 'Memberful_Metering_Access' )
  && Memberful_Metering_Access::get_current_decision( $lock_post_id ) === Memberful_Metering_Access::DECISION_ALLOW_SAMPLE ) {
  return false;
}
```

wordpress/wp-content/plugins/memberful-wp/src/content_filter.php
```diff
@@ -83,23 +83,43 @@ function memberful_wp_split_post_content_at_paywall_divider( $content ) {
...
+function memberful_wp_wrap_teaser_content( $content ) {
+  global $post;
+
+  /**
+   * Filters teaser content before the global teaser wrapper is rendered.
+   *
+   * @param string       $content Rendered teaser content.
+   * @param WP_Post|null $post    Current post, when available.
+   */
+  $content = apply_filters( 'memberful_teaser_content', $content, $post );
```
mrhead, line 100: Since the filter runs before the wrapper is added, the recipe preview ends up inside `.memberful-global-teaser-content`. The teaser CSS in `memberful_get_teaser_css()` fades `p:last-child`, so with a card appended the teaser paragraph loses its fade and a paragraph inside the card fades instead.

I think the card being inside the wrapper is fine, it just needs the fade to target it. Could we fade the wrapper's last child instead of its last paragraph? With a card the card fades out, without one the last paragraph fades as before.

```css
.memberful-global-teaser-content > :last-child {
    -webkit-mask-image: linear-gradient(180deg, #000 0%, transparent);
    mask-image: linear-gradient(180deg, #000 0%, transparent);
}
```

mrhead (commented):
Nice work, @av3nger!

I left a few suggestions closing some partial recipe leaks, and a few smaller
fixes. The leak fixes might seem overly cautious, but once we release this we want
to ensure there is no way to extract additional recipe information beyond what
should be visible to people without access.

Also, this PR is behind main, could you please merge it in?

av3nger:
@mrhead , I have addressed all the feedback from above.

wordpress/wp-content/plugins/memberful-wp/src/contrib/wp-recipe-maker.php
```diff
@@ -0,0 +1,925 @@
...
+
+  /**
+   * Register WP Recipe Maker and Memberful hooks.
+   */
+  private function register_hooks(): void {
+    add_filter( 'wprm_recipe_shortcode_output', array( $this, 'filter_recipe_shortcode_output' ), 100, 2 );
+    add_filter( 'wprm_recipe_snippet_shortcode_output', array( $this, 'filter_recipe_snippet_shortcode_output' ), 100, 3 );
+    add_filter( 'render_block', array( $this, 'filter_rendered_block' ), 100, 2 );
+    add_filter( 'do_shortcode_tag', array( $this, 'filter_shortcode_tag' ), 100, 3 );
+    add_filter( 'wprm_get_recipe_ids_from_post', array( $this, 'filter_recipe_ids_from_post' ), 10, 2 );
```
mrhead: Could we remove the `wprm_get_recipe_ids_from_post` and
`wprm_get_recipe_ids_from_content` filters? It looks like they are not needed,
since the outputs they guard are already covered by the
`wprm_recipe_metadata`, `wprm_recipes_on_page` and snippet filters.

Additionally, on WP Recipe Maker 10.8.0 they caused an issue with WP-Cron.
When it published a scheduled post with locked recipe cards, WPRM's
`save_post` handler rescanned the post with no logged-in user, the filter
returned an empty list, and WPRM deleted `wprm_parent_post_id` from the recipe.
This allowed anyone to access the full recipe via the print page, the REST API
and its own URL. On the latest version (10.8.1) WPRM added a permission check
that prevents the deletion of `wprm_parent_post_id` from recipes the current user can't edit, so the filters are redundant there.

mrhead:
Thanks @av3nger! The fixes look good.

Yesterday I forgot to post a comment about removing the recipe ID discovery filters, so I just did
it now: https://github.com/memberful/memberful-wp/pull/436#discussion_r3975668755

Also, it looks like main is still not merged in. It should fix the metering guard (see the thread on `memberful_metering_is_releasing()`, the `Memberful_Metering_Access` class only exists on main) and the listing excerpts of protected posts with a divider, which on this branch still show text from below the divider.

av3nger:
@mrhead , done

PR #437, opened 2026-09-01
Fix comments feed protection issues

mrhead:
* 24e8baaf **Fix comments feed protection issues**
  
  The comments feed was broken on sites with no protected posts because
  the feed query filter generated invalid SQL (`NOT IN ()`). The exclusion
  list also missed posts protected by term rules, so their comments
  leaked into the feed. Both feed paths now use
  `memberful_wp_user_disallowed_post_ids( 0 )`, which covers the full
  protection model for anonymous visitors (feeds are fetched by readers
  and crawlers without auth).
  
  Also guard against a null `$post` in `memberful_user_can_access_comments()`,
  which logged a PHP warning on feed requests for missing posts.
  
  Fixes: https://app.basecamp.com/3293071/buckets/5610905/card_tables/cards/10255383946

* d20dfba4 **Apply term protections to all post types**


mrhead:
Closing in favor of https://github.com/memberful/memberful-wp/pull/438.

PR #438, opened 2026-09-01
Fix comments feed protection issues

mrhead:
* 21397125 **Fix comments feed protection issues**
  
  - Fix broken SQL in the site-wide comments feed when no posts are protected
  - Fix PHP warning when checking comment access on feed requests for missing posts
  - Hide comments on term-protected posts from the site-wide comments feed
  - Check the singularity of the query being filtered instead of the main query
  - Reuse `memberful_can_user_access_post()` for the singular feed check instead
    of materializing the full list of protected post IDs
  
  Fixes: https://app.basecamp.com/3293071/buckets/5610905/card_tables/cards/10255383946

* a499ce1f **Apply term protections to all supported post types**

### DIFF
--- a/wordpress/wp-content/plugins/memberful-wp/src/comments_protection.php
+++ b/wordpress/wp-content/plugins/memberful-wp/src/comments_protection.php
@@ -33,11 +33,13 @@ function memberful_user_can_access_comments(){
 
   global $post;
 
-  // User has access to this post, we won't do anything.
-  if ( memberful_can_user_access_post( wp_get_current_user()->ID, $post->ID ) )
+  $post_id = isset( $post->ID ) ? (int) $post->ID : get_queried_object_id();
+
+  if ( ! $post_id )
     return true;
 
-  return false;
+  // User has access to this post, we won't do anything.
+  return memberful_can_user_access_post( wp_get_current_user()->ID, $post_id );
 }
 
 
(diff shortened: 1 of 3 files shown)


wordpress/wp-content/plugins/memberful-wp/src/comments_protection.php
```diff
@@ -71,9 +67,10 @@ function memberful_post_is_protected($post_id=null){
   if(!isset($post_id))
     $post_id=get_the_ID();
 
-  $acl= get_option( 'memberful_acl', array());
-  $restricted=memberful_get_protected_post_IDS();
-  return in_array($post_id, $restricted);
+  if ( empty( $post_id ) )
+    return false;
+
+  return ! memberful_can_user_access_post( 0, $post_id );
```
lmansur, line 73: Not blocking, but WDYT about defining a variable for `0`, like `anonymous_user_id` or something like that? Right now it's kind of a magic number until you read `memberful_can_user_access_post` and figure out what 0 does.

lmansur (approved):

wordpress/wp-content/plugins/memberful-wp/src/comments_protection.php
mrhead: Thanks for the suggestion, @lmansur!

I'm on the edge here. On one hand I agree that it might not be obvious.
On the other hand, `0` is a standard WordPress convention for logged out users
(`get_current_user_id()` returns `0` for anonymous visitors) and we use it in
`memberful_comment_feed_cwhere_filter()` as well, so I'm not sure extracting it
into a var everywhere is worth it.

With that said, I'll probably keep `0` here.

PR #447, opened 2026-09-17
MR-5 cache safe metering

av3nger:


js/metering.test.js
```diff
@@ -0,0 +1,386 @@
+const assert = require('node:assert/strict');
```
mrhead: This file gets included in the plugin zip.

Should we move it to tests/js/ and update the test script, or add /js/*.test.js to .distignore?

js/src/blocks/metering-countdown/render.php
```diff
@@ -10,8 +10,37 @@
...
+  || ( function_exists( 'memberful_metering_is_releasing' ) && memberful_metering_is_releasing( (int) get_the_ID() ) )
+) {
+  printf(
+    '<p %s hidden>%s</p>',
+    get_block_wrapper_attributes(
+      array(
+        'data-memberful-countdown'         => '1',
+        'data-memberful-template'          => wp_strip_all_tags( $template ),
+        'data-memberful-template-singular' => wp_strip_all_tags( $singular ),
+        'data-memberful-template-last'     => wp_strip_all_tags( $last_template ),
```
mrhead: In #450 I switched the countdown to escaping instead of stripping tags. Could we drop `wp_strip_all_tags()` here too? `get_block_wrapper_attributes()` already escapes the values. Thanks!

js/src/blocks/metering-countdown/render.php
```diff
@@ -10,8 +10,37 @@
...
+      )
+    ),
+    esc_html( wp_strip_all_tags( str_replace( '{count}', '', $template ) ) )
```
mrhead: Same here, `esc_html()` is enough.

PR #448, opened 2026-09-17
Improve the private member feed

mrhead:
Look the feed's member up with an explicit `meta_query` and reject an empty token before the query runs.


lmansur (approved):
Thanks for handling it.

